PR #27, opened 2013-12-26
CloseWithoutCloseableRule: support indirect implementation of Closeable

davidmc24:
Prior to this, the rule would fail if a class included a close method while either
extending a class that implements Closeable, or implementing an interface
that extends Closeable, either of which is a false positive.  The fix is to
use a slightly later compilerPhase, similar to CloneWithoutCloneableRule.
A new test method is included to provide test coverage for this case.


erdi:
This can probably now be closed, @chrismair.

Issue #94, opened 2015-05-23
Some tests incompatible with Java 8


topr:
The following three tests are failing if run with Java 8 JDK.

**UnsafeImplementationAsMapRuleTest**

- testMethodsMissingFromInheritedInterfaceViolate
```
java.lang.AssertionError: 
Expected message text: [Incomplete interface implementation. The following methods of FunnyIterator are not implemented by this map-to-interface coercion: [hasNext, makeLotsOfFun, remove]. Please note that calling any of these methods on this implementation will cause an UnsupportedOperationException, which is likely not intended.]
Found message text:    [Incomplete interface implementation. The following methods of FunnyIterator are not implemented by this map-to-interface coercion: [hasNext, makeLotsOfFun]. Please note that calling any of these methods on this implementation will cause an UnsupportedOperationException, which is likely not intended.]
```

- testMultipleViolations
```
java.lang.AssertionError: 
Expected message text: [Incomplete interface implementation. The following methods of java.util.Iterator are not implemented by this map-to-interface coercion: [hasNext, next, remove]. Please note that calling any of these methods on this implementation will cause an UnsupportedOperationException, which is likely not intended.]
Found message text:    [Incomplete interface implementation. The following methods of java.util.Iterator are not implemented by this map-to-interface coercion: [hasNext, next]. Please note that calling any of these methods on this implementation will cause an UnsupportedOperationException, which is likely not intended.]
```

- testSingleViolation
```
java.lang.AssertionError: 
Expected message text: [Incomplete interface implementation. The following methods of java.util.Iterator are not implemented by this map-to-interface coercion: [hasNext, remove]. Please note that calling any of these methods on this implementation will cause an UnsupportedOperationException, which is likely not intended.]
Found message text:    [Incomplete interface implementation. The following methods of java.util.Iterator are not implemented by this map-to-interface coercion: [hasNext]. Please note that calling any of these methods on this implementation will cause an UnsupportedOperationException, which is likely not intended.]
```

topr:
Hi @ArturGajowy :bowtie:

ArturGajowy:
@topr pull requests welcome ;>

topr:
Fixed by @ArturGajowy (I checked the build on Java8)

PR #95, opened 2015-05-23
Fix for #94: Some tests incompatible with Java 8 

ArturGajowy:
Changed the example interface used by tests from Iterator to Enumeration (which hasn't changed in JDK8 and is not likely to change soon)


topr:
:+1:

chrismair:
Thanks for the fix!

PR #97, opened 2015-05-25
Basic Gradle build setup with wrapper

topr:
Re-raising https://github.com/CodeNarc/CodeNarc/pull/96 against branch **gradle-migration**.

--

@chrismair 

- No changes to POM here.
- Groovy version is updated to the latest but only for Gradle build.
- Because of cyclic dependency between java and groovy sources joint compilation is configured - _compileGroovy_ task compiles sources for both languages. _compileJava_ task is executed as _compileGroovy_ depends on it but it does nothing.
BTW. maybe it's worth of considering to fix this dependency cycle?
- interesting thing to notice, the jar created is less than half of the size created by with Maven - class files are much smaller. Not sure why yet.
- adding wrapper implies committing one jar, it never changes though (probably you know that anyway)


chrismair:
Looks great. I will probably want to revert to Groovy 2.1.8 at some point since I am not ready to advance that yet, and I'd like to avoid introducing unintended dependencies on newer features. Thanks very much!

PR #104, opened 2015-06-02
Fix missing property in maven publishing config

topr:
...and you go like `./gradlew uploadArchives -PossrhUsername=christmar -PossrhPassword=plop`.

Or in your local $GRADLE_USER_HOME/gradle.properties set relevant properties like:
```
ossrhUsername=christmar
ossrhPassword=plop
```
No need for extra command line params then.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,11 +65,11 @@ uploadArchives {
             beforeDeployment { deployment -> signing.signPom(deployment) }
 
             repository(url: "https://oss.sonatype.org/service/local/staging/deploy/maven2/") {
-                authentication(userName: ossrhUsername, password: ossrhPassword)
+                authentication(userName: properties['ossrhUsername'], password: properties['ossrhPassword'])
             }
 
             snapshotRepository(url: "https://oss.sonatype.org/content/repositories/snapshots/") {
-                authentication(userName: ossrhUsername, password: ossrhPassword)
+                authentication(userName: properties['ossrhUsername'], password: properties['ossrhPassword'])
             }
 
             pom.project {


topr:
...and you go like `./gradlew uploadArchives -PossrhUsername=christmar -PossrhPassword=plop`.

Or in your local $GRADLE_USER_HOME/gradle.properties set relevant properties like:
```
ossrhUsername=christmar
ossrhPassword=plop
```
No need for extra command line params then.

chrismair:
I had those properties in my .gradle/gradle.properties, but this is even more flexible, Thanks.

topr:
You can still have them over there. It will pick up. Although that's not the best place for keeping local variables as the file is under version control.

Actually I haven't change it much. They could've been passed the same ways anyway. The problem was if not passed any build were failing with undefined property. Even build unrelated to releasing/uploading artifacts.
As we got here let me tell you there is tricky thing about gradle. Unless you know that yet, each task has configuration phase and execution phase. All our build.gradle for the moment contains only configurations. Configuration phase for a task is evaluated always, even if you run a different task while execution phase is evaluated only for a task(s) which are being run (specified on a command line or dependent).

Other thing is there are multiple ways of accessing gradle properties. Too many, I think. It may get confusing, especially given that some of them throw exception when the property is undefined and some just give null.

By the way, I used Groovy notation of accessing properties `Map` with square brackets and commas `properties['something']`. That's minor but wouldn't it be better to use simpler notation `properties.something`?

chrismair:
I meant that I had the properties in $GRADLE_USER_HOME/gradle.properties, and yes I understand your change will still support that.

I think I like the properties['something'] syntax better -- seems to make it stand out a bit more, but it is not a big deal.

Thanks for the heads up about the Gradle phases. I'll try to keep that in mind.

Issue #93, opened 2015-05-21
Some rules fail with NPE on traits


topr:
At least for the following rules:
- `BracesForClass`
- `SpaceAfterOpeningBrace`
- `SpaceBeforeOpeningBrace`

CodeNarc throws `NullPointerException` if applied on a Groovy `trait`. I guess the support for traits it's not implemented yet as it's quite new language feature. Although other rules I'm using work al lright which makes me think it shouldn't be to big chunk of work to do (how do you think?).

I'm running on:
- Groovy 2.4.3
- CodeNarc 0.23
- Gradle 2.4

Both Gradle 2.4 and 2.3 give the same error although with Gradle 2.3 stacktrace is printed:
```
$ ./gradlew clean codenarcMain
:buildSrc:compileJava UP-TO-DATE
:buildSrc:compileGroovy UP-TO-DATE
:buildSrc:processResources UP-TO-DATE
:buildSrc:classes UP-TO-DATE
:buildSrc:jar UP-TO-DATE
:buildSrc:assemble UP-TO-DATE
:buildSrc:compileTestJava UP-TO-DATE
:buildSrc:compileTestGroovy UP-TO-DATE
:buildSrc:processTestResources UP-TO-DATE
:buildSrc:testClasses UP-TO-DATE
:buildSrc:test UP-TO-DATE
:buildSrc:check UP-TO-DATE
:buildSrc:build UP-TO-DATE
:yolo:clean
:yolo:codenarcMainError from [org.codenarc.rule.formatting.SpaceBeforeOpeningBraceRule] processing source file [C:/dev/proj/yolo/src/main/groovy/AnyTrait.groovy]
java.lang.NullPointerException
	at org.codenarc.util.AstUtil.findFirstNonAnnotationLine(AstUtil.java:1085)
	at org.codenarc.rule.AbstractAstVisitor.sourceLine(AbstractAstVisitor.java:76)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
[...]

PR #108, opened 2015-06-13
Fix NPE when analyzing file with trait

alien11689:
It should fix issue https://github.com/CodeNarc/CodeNarc/issues/93


chrismair:
Thanks for the fix. I see that you upgraded to Groovy 2.4, understandably. I want to be careful to not introduce runtime dependencies on > 2.1, and your fix is in line with that, Thanks.

Issue #126, opened 2015-11-19
False positive UnnecessaryPublicModifier for groovy method with containing the word "public"


tomdcc:
Using 0.24.1

```groovy
class MyTest {

  void "some method with the word public in it"() {
    // etc
  }
}

Gives an `UnnecessaryPublicModifier` with the following message:
```
[SRC]void "some method with the word public in it"() {
[MSG]The public keyword is unnecessary for methods
```

ychulovskyy:
@chrismair this can be closed as fixed.

chrismair:
Thanks

PR #127, opened 2015-12-04
Fixed issue #126 False positive UnnecessaryPublicModifier for groovy …

ychulovskyy:
…method with containing the word "public"


chrismair:
Thanks for the contribution

Issue #119, opened 2015-09-03
BracesForIfElse with curly braces inside GString


Myllyenko:
CodeNarc 0.24.1 mistakenly marks statements of the following form as BracesForIfElse (sameLine = false) violation:
<code>if (someContainer."${SomeClass.SOME_CONSTANT}" != null)</code>

Myllyenko:
The same pseudo-violation detected in case of "if (someCollection.any { someClosure } )"

chrismair:
Fixed by #130

Myllyenko:
Could you please make a Codenarc realease in which this issue is fixed?

chrismair:
I'll try to get a release out within a few weeks.

PR #130, opened 2015-12-24
Fixed issue #119 BracesForIfElse with curly braces inside GString

ychulovskyy:


chrismair:
Nice work. Thanks for the fix!

Issue #115, opened 2015-08-07
UnusedImport does not handle Scala imports


uglyog:
Scala classes can names that end in dollars. CodeNarc UnusedImport rule always reports these as unused even when they are actually used.

The following code will violate this rule:

```groovy
import scala.None$

assert providerState == None$.empty()
```

ychulovskyy:
@chrismair The fix has been committed into master. Could you please close the issue?

PR #131, opened 2015-12-24
Fixed issue #115 UnusedImport does not handle Scala imports

ychulovskyy:


src/main/groovy/org/codenarc/rule/imports/UnusedImportRule.groovy
```diff
@@ -62,20 +64,31 @@ class UnusedImportRule extends AbstractRule {
...
+        // Define boundary as
+        //      right char is valid for class name
+        //      left char is not valid for class name
+        // or vice versa.
+        // Use boundary before and after class name.
+        // Yep, \b is much shorter :)
+        def regexp = /(?:(?=[a-zA-Z0-9_\$])(?<![a-zA-Z0-9_\$])|(?<=[a-zA-Z0-9_\$])(?![a-zA-Z0-9_\$]))/ +
+                /${pattern}/ +
+                /(?:(?=[a-zA-Z0-9_\$])(?<![a-zA-Z0-9_\$])|(?<=[a-zA-Z0-9_\$])(?![a-zA-Z0-9_\$]))/
+        (line =~ regexp).count
```
chrismair: That regular expression is pretty brutal.

I was able to achieve similar results with this:
        final INVALID = '[^a-zA-Z0-9_\\$]'
        def regexp = /($INVALID|^)${pattern}($INVALID|$)/

src/test/groovy/org/codenarc/rule/imports/UnusedImportRuleTest.groovy
```diff
@@ -162,12 +162,14 @@ class UnusedImportRuleTest extends AbstractRuleTestCase {
         final SOURCE = '''
             import java.io.InputStream
             import java.io.OutputStream
+            import scala.None$
```
chrismair: And I added another test case for catching a reference at the beginning of a line (see additional comment below):

    import scala.Other$

src/test/groovy/org/codenarc/rule/imports/UnusedImportRuleTest.groovy
```diff
@@ -162,12 +162,14 @@ class UnusedImportRuleTest extends AbstractRuleTestCase {
...
             import java.io.InputStream
             import java.io.OutputStream
+            import scala.None$
             import static Math.PI
             class ABC {
                 def run() {
                     String fff
                     InputStream input
                     OutputStream output
+                    None$.empty()
```
chrismair: This must be starting at the beginning of the line:
     
    Other$.value()

chrismair:
Merged manually. Thanks for the contribution!

PR #132, opened 2015-12-26
Fixed links in documentation

ychulovskyy:


ychulovskyy:
@chrismair Let me explain the reason of the change. If you open http://codenarc.sourceforge.net/codenarc-rule-index.html and click on UseAssertTrueInsteadOfAssertEquals it will open http://codenarc.sourceforge.net/codenarc-rules-junit.html but will NOT navigate to UseAssertTrueInsteadOfAssertEquals.
Expected result: navigate to UseAssertTrueInsteadOfAssertEquals rule at the opened page
The same with NestedBlockDepth

chrismair:
Thanks for the fix!

Issue #136, opened 2016-01-10
Idea about new rule: ArrayTrailingComma


ychulovskyy:
I've compared CodeNarc and CheckStyle rules https://docs.google.com/spreadsheets/d/1ZDapW9f-BlulvtWLY34oEtGZxqR4KbY65OPDfFZJ2Xk/edit?usp=sharing
There is one simple rule that we use in Java and it would be great to have it for Groovy.
It's http://checkstyle.sourceforge.net/config_coding.html#ArrayTrailingComma
I can try to implement it.

chrismair:
That's a good idea.

I would probably expect the CodeNarc rule to also apply to List and Map literals. In that case, TrailingComma might be a good rule name.

This should also be in the "convention" package/ruleset. Some teams will not want that rule. In fact, it would not shock me if somebody wanted an opposite rule.

Thanks.
Chris

ychulovskyy:
@chrismair Could you please close the issue since the changes have been merged into master?

PR #137, opened 2016-01-13
#136 Added TrailingComma rule (convention)

ychulovskyy:


chrismair:
Thanks for the contribution!

FYI... the StarterRuleSet* files are generated before a release, so no need to add new rules to those files.

PR #148, opened 2016-02-22
[DNM] Support Indent Checking

rahulsom:
This is the initial work on checking indents. This currently checks the initial indents for classes and methods. There is a lot more work to be done for this rule to be any use.

I would really appreciate any feedback as to whether I'm heading in the right direction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@
 build/
 src/main/java/META-INF/
 /.gradle
+/.shelf
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,7 @@
+
+#TODO: Sort the following line into the file
+- Indentation rule (formatting) - Checks if indents are proper
+
 CodeNarc Change Log
 -------------------------------------------------------------------------------
 http://www.codenarc.org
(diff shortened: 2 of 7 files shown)


chrismair:
Thanks for the contribution. That is an interesting approach. I think it is worth pursuing.

Without doing an exhaustive review, this is some initial feedback:

- Run all the tests, especially `RunCodeNarcAgainstProjectSourceCodeTest`. Note that the CodeNarc project uses spaces for indentation and the `NoTabCharacter` rule is enabled.
- Please rename the rule to "Indentation". 
- Set the rule Priority to 3 (since it is a non-functional issue)
- Default `useTabs` to false
- Remove @Override methods that only print debug and then call super
- Do not use `println`; use the `LOG` inherited from `AbstractRule`
- The names of the `indentSize` and `tabSize` properties may lead to confusion, especially when using spaces for indentation.

Some other cases to add tests for:

- Static Inner classes
- Anonymous inner classes
- Annotations on classes and methods
- Multi-line String literals, Map and List literals

Add your new rule to `RunCodeNarcAgainstProjectSourceCodeTest` (`RunCodeNarcAgainstProjectSourceCode.ruleset`) and then see what violations pop up. You can temporarily disable all other rules to get quicker feedback.

rahulsom:
Head up: Once I got to the part where I wrote tests that checked against comments, I realized this approach will not work. I'll close this PR.

PR #158, opened 2016-05-13
UnnecessaryConstructor: add rule to (not) ignore annotations

daspilker:
Empty constructors with annotations currently cause a UnnecessaryConstructor violation.

But some frameworks require annotations on otherwise empty constructurs, e.g. `@DataBoundConstructor` in [Stapler](https://github.com/stapler/stapler). Or sometimes it's necessary to mark an empty default constructor as deprecated.

This PR adds an option to UnnecessaryConstructor to allow empty constructors with annotations. It keeps the existing behavior as default.


chrismair:
Thanks for the contribution!

The _ignoreAnnotations_ flag is behaving opposite of what I would expect. In other rules, an "ignore" property typically indicates conditions where the rule ignores violations. I guess I was expecting _ignoreAnnotations_ = _true_ to skip over an empty constructor if the constructor had any annotations. 

Are you okay with changing the polarity of that flag? Or else rename the property it to make it more obvious what it is doing?

daspilker:
Makes sense, I changed the polarity of the flag.

chrismair:
Thanks for the contribution!

PR #160, opened 2016-05-20
Add Misordered Imports Rule

rahulsom:
This rule allows one to define an order for imports (including static imports) and then matches the imports against the rule.

The defaults are copied from IntelliJ IDEA.

It adds a dependency on `diffutils`. I'm not sure if that's acceptable. Please let me know if there's something I can do to make it less of a problem.

The logic for detecting this is fairly complicated, but seems to run in reasonable time.

There are lots of violations in the codebase detected by this rule.

Thanks!


chrismair:
Thanks for the contribution! Cool stuff! 

The implementation is quite complex. I am still working my way through understanding how it works. Some initial feedback:

-  I’d rather not introduce the dependency on **diffutils**, especially for such a narrow use case. Such dependencies can be a real headache down the road with future upgrades. Can you redesign the algorithm to remove that dependency?
- This will need a bunch more tests for coverage and confidence, including (but not limited to): invalid patterns, custom patterns – various combinations and permutations, no lines between groups; extra lines between groups; duplicate imports; static imports within group of regular imports and vice versa; comment lines interspersed within the imports; violations with aliases; any other edge cases you encountered while “fixing” other CodeNarc classes?
- The naming of the class as `Rule` is confusing since `Rule` is one of the main interfaces in the CodeNarc rule framework, also including their own “priority”.
- The method names `addViolationForImport()` and `addViolationForStaticImport()` were confusing and misnamed (I think).
- Do the `Rule `and `Import `classes need to be public?
- The help page (codenarc-rules-imports.apt) will need information on the _patterns_ configuration field, including explanation of valid entries in that list and what they mean, and what the ‘*’ means. Much of that same information should also be in javadoc or comments within the `MisorderedImportsRule`class.
- It may be simpler dealing with this if we treat adding this rule to the CodeNarc internal ruleset and “fixing” the 140+ CodeNarc classes as a separate, later pull request. That might make reviewing easier, plus there may be other pull requests for other issues in the interim, touching some of those files.

Thanks.

rahulsom:
Thanks @chrismair. I'll get those things addressed soon.

chrismair:
Nice work. I am glad you were able to remove that external dependency. I’m still working my way through it, but in the meantime …

- Add tests for **STATIC_LAST_PATTERN**, since that is part of the public API? 
- Change order of modifiers to **private static** rather than static private, and visibility comes first, so e.g. `private static abstract `and `private static final …`
- Make `addImport()`, `addStaticImport()` and `findMisorderedImports()` **private**, or **protected** where they need separate testing.
- Does the `findMisorderedImports()` method deserve its own class? Is it feasible to extract a class for the _Longest Common Subsequence_ algorithm, perhaps renaming `MisorderedImport `to just something like `Line`, to make it general-purpose? That could be separately testable, and would help make the current class not quite so large and complex. In any case, the logic in here deserves comprehensive tests all on its own.

rahulsom:
- [x] Add tests for STATIC_LAST_PATTERN, since that is part of the public API?

  - Added tests for `STATIC_LAST_PATTERN`.

- [x] Change order of modifiers to private static rather than static private, and visibility comes first, so e.g. private static abstractand private static final …

  - Done

- [x] Make `addImport()`, `addStaticImport()` and `findMisorderedImports()` `private`, or `protected` where they need separate testing.

  - They are now `private`

- [x] Does the `findMisorderedImports()` method deserve its own class? Is it feasible to extract a class for the Longest Common Subsequence algorithm, perhaps renaming `MisorderedImport` to just something like `Line`, to make it general-purpose? That could be separately testable, and would help make the current class not quite so large and complex. In any case, the logic in here deserves comprehensive tests all on its own.

  - Done

----

How do I check the generated documentation from `apt` files?

chrismair:
Thanks for the changes and enhancements.

- Please rename `LCSDiffUtil` to `LongestCommonSubsequenceDiffUtl`.
- Change the copyright years to 2016 in all file headers.
- The `addImport()` and `addStaticImport()` methods still seem to be public. 
- Refactor `applyTo()` to reduce size/complexity – i.e. extract private helper method(s). 
- The `Import` class is still public, and should be private.
- Change `abstract private static class AbstractImportPattern` to `private static abstract class AbstractImportPattern`.
- Are you still working on more tests? There are a few other error paths and edge cases mentioned above that need tests.
- `MisorderedImportsRuleTest`: Given the test method named `testSeparatorsViolationsWithStaticsLast()` has no violations, I do not understand what it is testing.
- Why is there a change (complete replacement) of `RunCodeNarcAgainstProjectSourceCode`?
- FYI… The project still uses Maven to generate the APT files: **mvn clean -f site-pom.xml site**

Thanks.
Chris

rahulsom:
- [x] Please rename `LCSDiffUtil` to `LongestCommonSubsequenceDiffUtil`.
  - Done
- [x] Change the copyright years to `2016` in all file headers.
  - Done
- [x] The `addImport()` and `addStaticImport()` methods still seem to be `public`.
  - Done
- [x] Refactor `applyTo()` to reduce size/complexity – i.e. extract `private` helper method(s).
  - Done
- [x] The `Import` class is still `public`, and should be `private`.
  - Done
- [x] Change `abstract private static class AbstractImportPattern` to `private static abstract class AbstractImportPattern`.
  - Done. That will be the next rule I'll begin to write, for sure :).
- [ ] Are you still working on more tests? There are a few other error paths and edge cases mentioned above that need tests.
- [x] `MisorderedImportsRuleTest`: Given the test method named `testSeparatorsViolationsWithStaticsLast`() has no violations, I do not understand what it is testing.
  - Sorry! That's now `testSeparatorsNoViolationsWithStaticsLast`.
- [x] Why is there a change (complete replacement) of `RunCodeNarcAgainstProjectSourceCode`?
  - The line endings were CRLF (Windows)  and now are LF (Unix + OS X).
    I've reverted it to CRLF. 
    I see that there are a lot of files that are CRLF and lot of them are LF.
    What's the standard for this code base?
- [x] FYI… The project still uses Maven to generate the APT files: `mvn clean -f site-pom.xml site`
  - Thanks!

chrismair:
Rahul: Are you still working on the tests and on the edge cases for imports?

chrismair:
I pulled down your changes and created a new **MisorderedImports** branch for CodeNarc (https://github.com/CodeNarc/CodeNarc/tree/MisorderedImports). I refactored the tests a bit, including renaming the tests to follow some quasi-conventions, but more importantly so that I could see which scenarios were tested.

`MisorderedImportsRuleTest`: I added tests for duplicate imports, extra blank lines between imports, comments in between imports, imports with aliases, and invalid patterns values. I’m sure there are more scenarios that are needed; the complexity of this rule scares me a bit.

Unexpected blank lines and extra blank lines between imports do not seem to matter. Is that intentional? That is not the behavior I would expect. 

Interspersing static imports in between the regular imports did not produce errors like I would have expected (see the tests).

Having comments in between the import statements seemed to work okay.

jlstrater:
Hi everyone! Is this PR still needed? If so, is anyone working on it or can I help?

dmurat:
It would be great to have this rule in CodeNarc. Are there any updates/plans about it?

src/main/groovy/org/codenarc/rule/imports/MisorderedImportsRule.groovy
```diff
@@ -0,0 +1,281 @@
...
+ *
+ *  import javax.acme.Switch
+ *  import java.acme.Cord
+ *  import java.toons.*
+ * </pre>
+ *
+ * <ul>
+ *     <li>The static imports are all on top, regardless of which package they belong to. And they're all sorted
+ *          alphabetically</li>
+ *     <li>Then there is a newline separating it from the next group of imports</li>
```
UlrichEckhardt, line 66: "a newline separating it from..." --> "an empty line separating them from...". These glitches repeats twice more below.

Just my personal taste would also be to add a period (".") after each list item to make them complete sentences, which they are otherwise.

src/main/groovy/org/codenarc/util/LongestCommonSubsequenceDiffUtil.groovy
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright 2016 the original author or authors.
```
UlrichEckhardt, line 2: dito

src/main/resources/codenarc-base-messages.properties
```diff
@@ -1,3 +1,8 @@
+
+# todo: manually sort your messages into the correct location
+MisorderedImports.description=Imports should be sorted alphabetically and in groups
+MisorderedImports.description.html=Imports should be sorted alphabetically and in groups
+
```
UlrichEckhardt, line 5: This file seems to use Windows line endings, apart from your addition. I'm not sure how to handle this best, but the result is inconsistent now.

src/site/apt/codenarc-rules-imports.apt
```diff
@@ -88,3 +88,45 @@ Imports Rules  ("<rulesets/imports.xml>")
 
     public class MyClass{}
 -------------------------------------------------------------------------------
+
+* {MisorderedImports} Rule
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  <Since CodeNarc 0.26>
```
UlrichEckhardt, line 94: This is surely not right, or?

src/test/groovy/org/codenarc/rule/imports/MisorderedImportsRuleTest.groovy
```diff
@@ -0,0 +1,204 @@
+/*
+ * Copyright 2016 the original author or authors.
```
UlrichEckhardt, line 2: dito

src/test/groovy/org/codenarc/util/LongestCommonSubsequenceDiffUtilTest.groovy
```diff
@@ -0,0 +1,130 @@
+/*
+ * Copyright 2016 the original author or authors.
```
UlrichEckhardt, line 2: dito

src/test/resources/RunCodeNarcAgainstProjectSourceCode.ruleset
```diff
@@ -42,6 +42,7 @@ ruleset {
 
     ruleset('rulesets/imports.xml') {
         exclude 'NoWildcardImports'
+        exclude 'MisorderedImports'
```
UlrichEckhardt, line 45: Move this up one line to keep the exclude directives sorted, too. ;)

CHANGELOG.txt
```diff
@@ -2,6 +2,10 @@ CodeNarc Change Log
 -------------------------------------------------------------------------------
 http://www.codenarc.org
 
+Changes in version 0.26.0 (??? 2016)
+--------------------------------------
+- #160: MisorderedImports rule (imports) - Imports should be sorted alphabetically and in groups
+
```
UlrichEckhardt, line 8: This is wrong. However, this file should perhaps be removed from the repo to avoid such errors in the future.

UlrichEckhardt (commented):
I generally like the idea a lot! There are a few glitches which I mentioned in the review and which should be easy to fix. I haven't tried this on a real code base yet, so I'm not commenting yet on how well it works. With luck, I'll get to it in the next days...

Thank you for this, Rahul Somasunderam!

PR #162, opened 2016-06-08
UnnecessaryPublicModifierRule should not include generics method

norbson:
Since groovy parser does not handle generics method without modifier in front, we are forced to add one.
Link to issue:
http://stackoverflow.com/questions/11395527/groovy-generics-failure


chrismair:
Good stuff. Thanks for the contribution!

norbson:
glad I could help! BTW. consider adding some short note about this exception in rule description.

PR #166, opened 2016-09-08
Exclude `javax.annotation.concurrent.Immutable` from GroovyLangImmuta…

mkobit:
…bleRule

This Rule would previously report that the Groovy annotation was being used when it was not.
This change makes it possible to use the `javax.annotation.concurrent.Immutable`.

----

## Testing

Ran `./gradlew check` locally


chrismair:
Nice work. Thanks for the contribution.

PR #183, opened 2016-12-28
Add ability to exclude files by name for NoDef rule

mathifonseca:
It's ok for Grails controller actions to be called something like `def index()`, so I'd like the NoDef rule to not apply in classes called `.*Controller`.

### DIFF
--- a/src/main/groovy/org/codenarc/rule/convention/NoDefRule.groovy
+++ b/src/main/groovy/org/codenarc/rule/convention/NoDefRule.groovy
@@ -34,15 +34,19 @@ class NoDefRule extends AbstractRule {
     int priority = 3
     protected static final String MESSAGE = 'def should not be used'
     String excludeRegex
+    String excludeFilesRegex
 
     @Override
     void applyTo(SourceCode sourceCode, List<Violation> violations) {
         Pattern excludeFilter = excludeRegex ? ~/.*def\s+$excludeRegex.*/ : null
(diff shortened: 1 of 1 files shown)


chrismair:
This capability should already be available using the _doNotApplyToFilesMatching_  or
_doNotApplyToFileNames_ standard properties available on all rules:
http://codenarc.sourceforge.net/codenarc-configuring-rules.html#Standard_Properties_for_Configuring_Rules

mathifonseca:
Thanks @chrismair! I didn't know about that option and only read the docs [here](http://codenarc.sourceforge.net/codenarc-rules-convention.html#NoDef) that talk about the exclusion by ClassName. Should be nice to add the doNotApplyToFileNames option in the docs maybe?

chrismair:
I'll make a note about that. I don't want to have to cut and paste that info into 300+ places, so maybe a link at the top of each rules category page. Thanks.

mathifonseca:
I think in my case it would be enough to add something about the `doNotApplyToFileNames` property only to the rules where `doNotApplyToClassNames` is not applicable. Thanks very much for the support!

chrismair:
For what it's worth, I added a note and link to common rule config properties from each rule to the rules where *applyToClassNames* and *doNotApplyToClassNames* properties are not available

mathifonseca:
That's great, thanks!

PR #196, opened 2017-03-25
fix the false positive that occurs when a closure is the last item in a map entry

### LINKED ISSUE #194: False positive for SpaceAfterClosingBrace for method call expression
The following code causes a violation:

```
[data: relations.collect { Relation.of(it) }]
```

even with `checkClosureMapEntryValue` set to `true`.

jlstrater:
resolves #194


chrismair:
Nice solution. Thanks for the contribution.

PR #198, opened 2017-04-09
fix false positive for trailing comma when sending method parameters by name

jlstrater:


chrismair:
Thanks for the contribution!

PR #199, opened 2017-04-09
add new rule to check for simple cases where 3 or more ifs could be rewritten as a switch statement

jlstrater:


src/main/groovy/org/codenarc/rule/convention/CouldBeSwitchStatementRule.groovy
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright 2015 the original author or authors.
```
chrismair: Change copyright year to 2017

src/main/groovy/org/codenarc/rule/convention/CouldBeSwitchStatementRule.groovy
```diff
@@ -0,0 +1,108 @@
...
+class CouldBeSwitchStatementRule extends AbstractAstVisitorRule {
+
+    String name = 'CouldBeSwitchStatement'
+    int priority = 3
+    Class astVisitorClass = CouldBeSwitchStatementAstVisitor
+    String errorMessage = 'Code could use switch statement'
+}
+
+class CouldBeSwitchStatementAstVisitor extends AbstractAstVisitor {
+    BinaryExpression prev = null
```
chrismair: Please make these 3 private fields instead of (public) properties

src/test/groovy/org/codenarc/rule/convention/CouldBeSwitchStatementRuleTest.groovy
```diff
@@ -0,0 +1,126 @@
...
+            } else if (!x && i) {
+                doAnotherThing()
+            }
+        '''
+        assertViolations(SOURCE,
+            [lineNumber:2, sourceLineText: 'if (x == 1) {', messageText: rule.errorMessage],
+            [lineNumber:12, sourceLineText: 'if (y instanceof Integer) {', messageText: rule.errorMessage],
+            [lineNumber:23, sourceLineText: 'if (p.value instanceof Integer) {', messageText: rule.errorMessage])
+    }
+
```
chrismair: Please add separate test cases for non-matching conditions, either indicated in separate test method names, or else by inline comments within the test method(s). I know you have a few in the above test. Conditions include (but are not limited to):

- different left expression types
- mixed left expression types (some supported, some not)
- unsupported operation types
- mixed operation types (some supported, some not)
- variations of` PropertyExpression` left expression types, which seem to be the most complicated (e.g. matching _objectExpression_ but not _property_ and vice versa)
- same expression type and/or operation but across separate classes
- two matching expressions (does not reach the threshhold of three)

jlstrater: Thank you! I think I have all of them now. The only one I'm really not sure about is what you mean by `same expression type and/or operation but across separate classes`. I think it might mean this: https://github.com/CodeNarc/CodeNarc/pull/199/commits/037c3bb6c90ade00818c0a1ea3d2799f35f0324d#diff-162de04ee2e56a411f15b002c7b63285R175 ?

chrismair: Thanks for the (well-named) tests. And yes, the multi-class test is what I meant.

chrismair:
Thanks for the well-crafted contribution!

jlstrater:
It looks like this was closed. :( Was it supposed to be merged?

chrismair:
Sigh, yes, I must have accidentally closed instead of merge .. I was wondering why the changes didn't show up when I updated my local :)

PR #200, opened 2017-04-09
fix the error message in the logs from the space after closing brace rule

jlstrater:
Unfortunately, I'm a little stumped by this one.  The errors in the logs like the one below: 
screen shot 2017-04-09 at 5 50 58 pm
come from `RunCodeNarcAgainstProjectSourceCodeTest`.  Since the errors are only in the logs and they aren't violations, it didn't fail in the initial pull request.  I cannot replicate it in the tests even using the same source code from the files it complains about in the logs.  However, this fix seems to resolve the messages.  I would appreciate feedback especially if someone can figure out how to write a failing test for it.


chrismair:
Thanks for another contribution!

Issue #197, opened 2017-03-27
Support @SuppressWarnings on local variables


mkobit:
I have a use case where I need to suppress a CodeNarc rule in a script. I would like to be able to do something like:

```groovy
@SuppressWarnings('UnusedVariable')
final myVar = thing.findVar()
```

I am using `0.26.0` from Gradle and would like to use my normal ruleset in general, with possible some disabled in scripts, but make it visible in those scripts rather than with with `codenarc.groovy` configuration I am using.

Relevant discussion from before: https://sourceforge.net/p/codenarc/mailman/message/28420205/

chrismair:
For what it's worth, the current design of the **@SuppressWarnings** handling does not include variables:

> You can use the @SuppressWarnings annotation on a class, method, constructor, field or property to suppress one or more CodeNarc rules within its source code.

I am not saying it is impossible to enhance that, but just that it was an intentional limitation.

mkobit:
Good point! I didn't see that originally, but thanks for highlighting it.

For reference of tohers that may come to this issue, I found the documentation you quoted at http://codenarc.sourceforge.net/codenarc-configuring-rules.html

mkobit:
Updated the title to better reflect the request.

aadamovich:
Submitted a pull request that should make @SuppressWarnings work on local variables

chrismair:
Fixed by #206

PR #206, opened 2017-06-01
fix for issue #197

aadamovich:


src/main/groovy/org/codenarc/analyzer/SuppressionAnalyzer.java
```diff
@@ -97,18 +99,27 @@ private void init() {
...
+                        populateLineNumbers(expression, result, numLines, ruleName);
+                    }
+                    super.visitDeclarationExpression(expression);
+                }
+            };
+            for (MethodNode methodNode : from(classNode.getMethods())) {
                 for (String ruleName : getSuppressedRuleNames(methodNode)) {
                     populateLineNumbers(methodNode, result, numLines, ruleName);
                 }
+                declarationVisitor.visitMethod(methodNode);
```
mkobit: Would it be a good idea to also apply this to constructors?

aadamovich: Good point!

chrismair:
Great contribution. Thank you!

Issue #207, opened 2017-06-05
ParameterCount should ignore methods with @Override annotation


ychulovskyy:
Problem:
Sometimes we have to use 3rd party library that has method with many parameters.
We don't want to increase maxParameters and don't want to ignore the rule in many classes.

Possible solution:
add ignoreOverriddenMethods parameter to ignore methods with @Override annotation?

Checkstyle already has such feature http://checkstyle.sourceforge.net/config_sizes.html#ParameterNumber
I can prepare pull request if you approve the feature.

chrismair:
I like that. Sounds good.

ychulovskyy:
@chrismair shall we close the issue?

chrismair:
yes, thanks.

PR #208, opened 2017-06-06
#207 - ParameterCount should ignore methods with @Override annotation

ychulovskyy:


src/main/groovy/org/codenarc/rule/size/ParameterCountRule.groovy
```diff
@@ -17,8 +17,8 @@ package org.codenarc.rule.size
 
 import org.codehaus.groovy.ast.ConstructorNode
 import org.codehaus.groovy.ast.MethodNode
-import org.codenarc.rule.AbstractAstVisitorRule
 import org.codenarc.rule.AbstractAstVisitor
+import org.codenarc.rule.AbstractAstVisitorRule
```
ychulovskyy: Formatted with default IntelliJ Idea formatter

src/test/groovy/org/codenarc/rule/size/ParameterCountRuleTest.groovy
```diff
@@ -189,6 +189,65 @@ class ParameterCountRuleTest extends AbstractRuleTestCase {
...
+            class TestClass {
+
+                TestClass() {
+                }
+
+                void someMethod() {
+                }
+
+                @Override
+                void someMethod(int arg1, int arg2, int arg3, int arg4, int arg5) { ${violation('method TestClass.someMethod')}
```
chrismair: Can you remove the ${violation... ?  It is confusing.

chrismair:
Thanks for the contribution!

chrismair:
I changed *ignoreOverriddenMethods* to default to true. I think that in most cases that is reasonable default behavior. And since that is reducing the number of violations rather than adding new ones, it should not cause *breaking changes* for anyone.

Issue #203, opened 2017-05-29
Consider switching from log4j to slf4j


mkobit:
CodeNarc has a dependency on [`log4j:log4j:1.2.14`](https://github.com/CodeNarc/CodeNarc/blob/master/build.gradle#L25). This forces library consumers to take on a logging implementation themselves when they may want to provide their own.

It would be useful for consumers to swap out the current usage of log4j and replace it with slf4j.

erdi:
While there is a point in your suggestion this would be a breaking change and a lot of work for relatively little benefit (it's "just logging" in the end).

There is a simple workaround for you as a consumer though, so if you feel really strongly about it I would suggest you use [`log4j-over-slf4j`](https://www.slf4j.org/legacy.html#log4j-over-slf4j).

mkobit:
Thanks for the quick response! Good suggestion with using `log4j-over-slf4j`, that will probably be what I move to do.

I agree that it is a breaking change that does not offer much benefit (in the near term at least) and would almost certainly introduce a bunch of problems for existing consumers.

Going to close this since I think the workaround you provided is a good approach. This may be a change to consider somewhere down the line, but not worth pursuing right now due to the reasons you brought up.

chrismair:
Other issues have been opened re: CodeNarc and Log4J 1.x: #211 and #214 .

I understand that the dependency on Log4J 1.x gets more problematic, or at least annoying, as we get farther and farther beyond its end-of-life. I am open to upgrading to either Log4J 2.x or to [SLF4J](https://www.slf4j.org/).

I am leaning toward SLF4J. Any other considerations or strong opinions?

erdi:
I'm in favour of going with SLF4J because with it you write your logging code against a simple, clearly defined API and let your consumers plug in logging implementation of their choice.

M-r-A:
Hmm... If SLF4J allow me to plug in LOG4J 2.x, why not?
And i want the same thing with GMetrics, like in https://github.com/thetric/gmetrics by @thetric. 
Also, it is possible to recheck and update other dependencies in both projects in process of upgrade.
Can you, @chrismair?

chrismair:
I have added a GMetrics issue to switch to SLF4J:
https://sourceforge.net/p/gmetrics/feature-requests/34/

Are there specific other dependencies you think need to be updated?

Fedos86:
It's very good idea to use SLF4J.

M-r-A:
@chrismair
[...]

PR #211, opened 2017-06-12
WIP: Log4j2/GMetrics upgrade

thetric:
**DO NOT MERGE!1**

The Log4j 1 support has terminated quite [a long time ago](https://blogs.apache.org/foundation/entry/apache_logging_services_project_announces). This PR updates the logging facilities to the latest version Log4j2.

A side effect is that I've replace the original GMetrics dependency with [my fork](https://github.com/thetric/gmetrics) that already uses log4j2 instead of log4j.

This PR is currently WIP to get feedback from the maintainers/contributors.


chrismair:
Hmmm, if I were to upgrade logging, I'd probably be more inclined to switch to logback/slf4j.

And this was discussed in #203. Perhaps continuing that discussion would be appropriate if you feel strongly about this.

Thanks for the suggestion and work, in any case.

chrismair:
FYI... I have added a GMetrics issue to switch to SLF4J:
https://sourceforge.net/p/gmetrics/feature-requests/34/

And also one to switch from Maven to Gradle:
https://sourceforge.net/p/gmetrics/feature-requests/33/

At the very least your fork will provide some help in the migration. Thanks again.

PR #213, opened 2017-07-20
Don't run BracesForClassRule against script ClassNodes

SimonStJG:
Scripts do not have opening and closing braces, so in rare situations this
check will backfire and report a violation of this rule, for example when
run against the following script:

def x = """
{
}
"""


chrismair:
Thanks for the contribution!

PR #219, opened 2017-08-13
Use the right classloader for loading AST transformations when compiling analysed code.

### LINKED ISSUE #218: Incorrect class loader is used for loading AST transformations when compiling analysed code
It took me a while but I managed to get [support for CodeNarc Ant task's classpath property added to Gradle's CodeNarc plugin](https://github.com/gradle/gradle/pull/2325). Today I got around to testing it on one of my projects only to realise that there are still some issues with compiling classes for analysis by enhanced rules.

I tracked it down to the following compilation error message:

> Not an ASTTransformation: org.codehaus.groovy.transform.EqualsAndHashCodeASTTransformation declared by groovy.transform.EqualsAndHashCode

A little bit of googling revealed the following historical Groovy issue: https://issues.apache.org/jira/browse/GROOVY-5044. Cédric's comment on stating that

> the "transform class loader" used by the compilation unit (that is to say the compiler) must, at least, be the same or a child classloader of the classloader which loads groovy classes

got me looking at how compilation is performed in `AbstractSourceCode.init()`. Currently the `CompilationUnit` created in that method does not get any `transformLoader` passed which means the default of context classloader is used. That classloader is not the one that loads Groovy compiler (Groovy compiler is loaded by the classloader which loads CodeNarc classes) causing [this condition](https://github.com/apache/groovy/blob/c1170ef867e4487b64362769108c084426ace46e/src/main/org/codehaus/groovy/transform/ASTTransformationCollectorCodeVisitor.java#L199) to always fail.

A PR to follow.

erdi:
Closes #218.

Notes:
1. I had to create a submodule for the Gradle integration test because of conflicting Groovy versions. After upgrading to GMetrics to `1.0` CodeNarc is now depending on and compiled against Groovy `2.4.12` (even though the build is specifying the version as `2.1.8` in `gradle.properties`) but using Gradle Test Kit from Gradle `4.0` (current version used by the build) forces Groovy version to be `2.4.11`. I didn't want to to force a different version of Groovy when running all tests compared to what is used for compilation hence I created a separate submodule which is using `2.4.11` for that single Gradle integraiton test. We could update to Gradle 4.1 which should make Groovy required by the Test Kit to be `2.4.12` which in turn would allow us to get rid of that submodule but I'm not sure we want to tie Groovy version used to compile CodeNarc with the one used by Gradle for its runtime.
1. Support for CodeNarc Ant task's `classpath` property has not been yet officially released in Gradle hence the integration test is using a nightly build, namely `4.2-20170813022644+0000`. As I've mentioned in a comment in that test, we should probably change that test to use `4.2` when it's released.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@
 /hamlet_docs
 /*.patch
 /*.iws
-/*.iml
+*.iml
 /*.ipr
 /AntBuilderTestXmlReport.xml
 /*.html
@@ -20,3 +20,4 @@ build/
 src/main/java/META-INF/
 /.gradle
 /classes
+/out
--- a/integration-test/build.gradle
+++ b/integration-test/build.gradle
@@ -1,0 +1,25 @@
+plugins {
+    id 'groovy'
+    id 'idea'
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    testCompile project(':')
+    testCompile gradleTestKit()
+}
+
+configurations.all {
+    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
+        if (details.requested.group == 'org.codehaus.groovy' && details.requested.version == '2.4.12') {
+            details.useVersion '2.4.11'
+        }
+    }
+}
+
+test {
+    systemProperty 'codenarc.test.projectPath', rootProject.projectDir.absolutePath
+}
--- a/integration-test/src/test/groovy/org/codenarc/gradle/GradlePluginCompilationClasspathTest.groovy
+++ b/integration-test/src/test/groovy/org/codenarc/gradle/GradlePluginCompilationClasspathTest.groovy
@@ -1,0 +1,92 @@
+package org.codenarc.gradle
+
+import org.gradle.testkit.runner.BuildResult
+import org.gradle.testkit.runner.GradleRunner
+import org.junit.Rule
+import org.junit.Test
+import org.junit.rules.TemporaryFolder
+
+class GradlePluginCompilationClasspathTest {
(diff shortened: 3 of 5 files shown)


erdi:
Is there anything I can do to help in getting this merged?

chrismair:
My bad. This slipped through the cracks. I'll take a look this weekend.

erdi:
No worries. Good that I pinged you. :)

chrismair:
So if I set `transitive = false` for the GMetrics dependency, that restores CodeNarc to using the Groovy 2.1.8 version. Does that address or help with item #1 you mentioned?

erdi:
While modifying the dependency on GMetrics will allow for the version specified in `gradle.properties` to be used (i'd probably personally use `exclude group: 'org.codehaus.groovy'` instead of `transitive = false`), it does not help with the first item I mentioned. 

The problem is that by depending on `gradleTestKit()` for `testCompile` configuration it pulls in the version of Groovy used by the Gradle version used to run the build. I don't seem to be able to find a way to override the version used by `gradleTestKit()` and if any other version is defined in the configuration then two groovy versions are put on the classpath and an error like this occurs:

> groovy.lang.GroovyRuntimeException: Conflicting module versions. Module [groovy-all is loaded in version 2.1.8 and you are trying to load version 2.4.11

So it's either putting that test into a separate submodule or tying Groovy version for compiling and running the tests to the version used by the Gradle version used to run the build.

I think it's best and safest to keep this test as a separate submodule.

chrismair:
Thanks for the contribution and for the extra research!

PR #220, opened 2017-08-14
215 - New rule: UnnecessarySetter

ychulovskyy:


src/main/groovy/org/codenarc/rule/generic/StatelessClassRule.groovy
```diff
@@ -82,7 +82,7 @@ class StatelessClassRule extends AbstractAstVisitorRule {
      *      include wildcard characters ('*' or '?'). Any names specified here are joined to
      *      the existing <code>ignoreFieldNames</code> property value. 
      */
-    void setAddToIgnoreFieldNames(String moreFieldNames) {
```
chrismair: This method is deliberately named `setAddToIgnoreFieldNames()` so that it can be used as a *addToIgnoreFieldNames* property in **codenarc.properties**, e.g. `addToIgnoreFieldNames = foo, bar`. Renaming it as you have will break that, so please revert that change.

src/test/groovy/org/codenarc/ant/CodeNarcTaskTest.groovy
```diff
@@ -91,7 +91,7 @@ class CodeNarcTaskTest extends AbstractTestCase {
     void testExecute_SingleRuleSetFile() {
         def codeNarcRunner = createAndUseFakeCodeNarcRunner()
 
-        codeNarcTask.addConfiguredReport(new Report(type:'html', toFile:HTML_REPORT_FILE))
```
chrismair: For better or worse, in CodeNarc we have adopted the convention of no spaces after the colon for map entries, e.g. `[a:b]`. Please stick to that where feasible and revert all of the corrections you made like that in the tests.

ychulovskyy: @chrismair Sure, will fix till EOW

chrismair:
Are you still planning on refining the pull request? If not, I can just pull in the new rule and ignore those other changes?

ychulovskyy:
@chrismair Sorry for delay, I have a lot of events.
Let me finish it till Tuesday.

ychulovskyy:
@chrismair test passes locally but fail on travis-ci.
I don't see details from trevis-ci, just
*****************************
 :test FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':test'.
> Process 'Gradle Test Executor 2' finished with non-zero exit value 137
********************************
Do you have access to full logs?
Does it passes on your local machine (maybe it's travis configuration issue)?

chrismair:
The Travis CI failure (137) was an out-of-memory error; I had to tweak the Travis Ci config file to resolve that.

Thanks for the contribution!

chrismair:
FYI... I downgraded from Groovy 2.4 to 2.3. That required a minor change to the **UnnecessarySetterRule** code; it dumbs down the violation message a bit, no longer including the assignment text. See #228 for commit info.

PR #232, opened 2017-09-13
update link to the gradle maven plugin

jlstrater:


jlstrater:
Ok. It's not just me then.  I had to change the groovy version in the gradle properties file to get the project to build. Is that expected?

chrismair:
I'll take a look. I think it is due to the new integration-test submodule for testing different classpaths.

erdi:
This line needs changing after changing the version of Groovy used by the
root project:
https://github.com/CodeNarc/CodeNarc/blob/master/integration-test/build.gradle#L18.
Have the test pass for you locally Chris before you pushed?

Actually, thinking of it now, it's probably better to exclude that group
from the dependency on root project (
https://github.com/CodeNarc/CodeNarc/blob/master/integration-test/build.gradle#L11)
and remove the whole dependency substitution block (
https://github.com/CodeNarc/CodeNarc/blob/master/integration-test/build.gradle#L16).
That way we won't have to change it that much.

I'll submit a PR when I'm back from holidays next week if you don't get
round to it yourself by then.

chrismair:
Marcin: If you could do that PR, I'd appreciate it. Thanks.

I could have sworn I ran the full `gradle test` from the command-line before the release, but I guess I did not. I typically just run the tests from within IDEA.

erdi:
@chrismair I've just created https://github.com/CodeNarc/CodeNarc/pull/233 to address this.

chrismair:
Thanks Marcin.

Jenn: Thanks for the contribution!

chrismair:
I have published the updated page to the CodeNarc web site.

PR #233, opened 2017-09-19
Change dependency substitutions to exclusions in integration-test module

erdi:
This ensures that the build does not need to be modified every single time the versions of groovy or gradle used are changed


chrismair:
Thanks!

PR #239, opened 2017-09-30
Ignore super calls in UnnecessarySetterRule

### LINKED ISSUE #234: UnecessarySetter should ignore calls on superclass
`UnecessarySetter` rule reports violations on calls like:

```super.setValue(2)```

but

```super.value = 2```

does not compile, therefore the rule should ignore any calls on the superclass.

erdi:
Fixes #234


chrismair:
Nice. Thanks for the contribution!

erdi:
My pleasure!

PR #242, opened 2017-10-14
Removing duplicate gradle property settings.

russellsanborn:
`group` and `version` are both defined in build.gradle already.


chrismair:
Thanks!

PR #243, opened 2017-10-14
TrailingComma Rules - Allowing inline Lists in Annotations

russellsanborn:
I was having false CodeNarc errors show up for the TrailingComma Rule with code similar to this:
```groovy
@ToString(
    ignoreNulls = false,
    excludes = ['bear', 'raccoon']
)
class TestClass{}
```

Digging through the code, the list was marked as starting with `excludes = [` and ending with `)` instead of ending with `]`. So when passing the expressions in to `sourceLinesBetween()`, the returned nodes would be `[']', ')']` (I trimmed the whitespace to make it easier to see). Luckily the lastExpression in these single line Lists has the same `lineNumber` and `lastLineNumber`. Awesome! Unfortunately that is not the same for Maps.

For List checks, I found that you don't actually need to check for the trailing comma. The following code was sufficient and all tests pass. However, I figured you wanted the consistency with the Map detections, and the same style does not work for maps unfortunately.
```groovy
 @Override
 void visitListExpression(ListExpression expression) {
     if (expression.expressions.isEmpty()) {
         return
     }
     if (rule.checkList && !lastExpressionIsEndOfExpression(expression.expressions[-1])
     ) {
         addViolation(expression, 'List should contain trailing comma.')
     }
 }
```

### DIFF
--- a/src/main/groovy/org/codenarc/rule/convention/TrailingCommaRule.groovy
+++ b/src/main/groovy/org/codenarc/rule/convention/TrailingCommaRule.groovy
@@ -56,7 +56,10 @@ class TrailingCommaAstVisitor extends AbstractAstVisitor {
         if (isOneLiner(expression) || expression.expressions.isEmpty()) {
             return
         }
-        if (rule.checkList && !hasTrailingComma(expression.expressions[-1], expression)) {
+        if (rule.checkList &&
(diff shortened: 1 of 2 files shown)


chrismair:
Thanks for the research and the contribution!

russellsanborn:
Thanks for CodeNarc!

PR #248, opened 2017-10-28
Add MissingOverrideAnnotationRule.

### LINKED ISSUE #247: New rule to detect missing @Override annotations
A rule that checks for methods that override a method in a superclass or implement a method in an interface but are not annotated with `@Override`. 

Rationale behind the rule comes from Joshua Bloch's Effective Java 2nd Edition, Item 36: Consistently use the `Override` annotation.

erdi:
Closes #247.

Please let me know if you would like another PR with this new rule enabled in `RunCodeNarcAgainstProjectSourceCode.ruleset`, @chrismair. I've checked and it causes ~500 violations but I'm more than happy to work through fixing them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 TBD: Version 1.1 (??? 2018)
 --------------------------------------
+New Rules
+   - #247: MissingOverrideAnnotation rule (enhanced) - Checks for methods that override a method in a super class or implement a method in an interface but are not annotated with @Override. (Marcin Erdmann)
+
 Updated/Enhanced Rules and Bug Fixes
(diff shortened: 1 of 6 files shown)


chrismair:
That looks good. Can you add test(s) for inner and/or anonymous classes, to be sure those are handled properly?

erdi:
Good point. I amended the commit with some tests around inner and anonymous classes.

chrismair:
Good stuff!. Thanks for the contribution.

If you would *like* to do another PR to enable that rule in the CodeNarc test ruleset, then I'd be happy to pull that in. But I am also fine if you want to leave that out.

Issue #224, opened 2017-08-24
CompareToWithoutComparable


jameskleeh:
This rule throws an error in the case that an interface the class implements extends Comparable.

Example:
```
interface Bar extends Comparable<Bar> {}

class Foo implements Bar {
  int compareTo(Bar o) {}
}
```

chrismair:
Unfortunately, the rule is not able to determine the full information for super-classes or super-interfaces, because the AST does  not provide that information (during the compiler phase at which CodeNarc operates, by default). I will add a note to the rule description on the website, about this being a known limitation.

Your options include disabling the rule overall or for specific classes, or @SuppressWarnings on specific usages.

erdi:
I think we have the following options here:
1. If we care about backwards compatibility then we can introduce an enhanced rule with similar functionality and leave the current rule as is with a note in docs about it's deficiencies and a ointer to the new rule.
2. Change the rule in a backwards incompatible way to become an enhanced rule and support the usecase in this issue.

chrismair:
I agree with the options you listed. I have been conflicted about how to exploit the enhanced rule capabilities. I want to avoid breaking existing users, so I don't think I would be in favor of option 2. 

I wondered whether it was feasible to optionally extend existing rules, based on a config/system property or based on dynamic classpath analysis, to perform the _enhanced_ processing only when enabled. But that is probably too complicated or fragile.

erdi:
I do not think that using a configuration based approach is complicated or fragile. I don't see any problem with introducing a property on rules like that, for example `enhancedMode` that is `false` by default but can be changed to be `true` either manually in config for the rule or by setting, say `org.codenarc.enhancedModeForDualRules`, system property. The rules in question could then use a different `compilerPhase` and AST introspection logic depending on the value of the config property.

PR #249, opened 2017-10-28
Add enhanced mode to CompareWithoutComparable

erdi:
To avoid false positives when a superclass implements `Comparable`.

If you are happy with this we can take it to the next level in a separate PR by adding support for changing the default value of `enhancedMode` via a system property as suggested by me in #224.

### DIFF
--- a/src/main/groovy/org/codenarc/rule/design/CompareToWithoutComparableRule.groovy
+++ b/src/main/groovy/org/codenarc/rule/design/CompareToWithoutComparableRule.groovy
@@ -16,6 +16,7 @@
 package org.codenarc.rule.design
 
 import org.codehaus.groovy.ast.MethodNode
+import org.codehaus.groovy.control.Phases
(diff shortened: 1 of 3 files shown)


chrismair:
I like it! Looks good. Thanks.

If you do move forward with the system property, what do you think of just "org.codenarc.enhancedMode"?

erdi:
Naming things is hard, especially when English is not your first language,
like in my case. So if you are willing to do it for me by providing a
suggestion then I won’t complain. :) org.codenarc.enhancedMode shall it be!

Btw, can you think of any other rules that would benefit from enhanced
mode? I would like to introduce a base class for them before I go ahead
with implementing of switching the behaviour globally using a system
property.

chrismair:
I have not done a deep examination, but other rules that might be candidates for "enhanced" enhancements include:
- CloseWithoutCloseable
- UnusedPrivateMethod

PR #251, opened 2017-10-28
Remove net.researchgate.release plugin from build

### LINKED ISSUE #250: Remove net.researchgate.release plugin from build
As far as I understand it's not used but it causes the following warning when running the build:

> The Task.leftShift(Closure) method has been deprecated and is scheduled to be removed in Gradle 5.0. Please use Task.doLast(Action) instead.

erdi:
Closes #250

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,6 @@ plugins {
     id 'maven'
     id 'signing'
     id 'idea'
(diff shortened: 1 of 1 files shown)


chrismair:
I meant to do this before the last release, but forgot. Thanks for the contribution, as always!

PR #253, opened 2017-11-03
Added some consistency to getting dependency version

russellsanborn:
I added all dependency versions to the gradle.properties file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,6 +5,13 @@ plugins {
     id 'idea'
 }
 
+ext {
+    commonsCliVersion = '1.4'
+    junitVersion = '4.12'
+    GMetricsVersion = '1.0'
+    slf4jVersion = '1.7.25'
+}
+
 group = "org.codenarc"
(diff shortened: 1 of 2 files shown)


chrismair:
Thanks for the suggestion and work to implement it.

But I think that just makes the "build.gradle" harder to read/maintain. I am more inclined to move *groovyVersion* and *slf4jVersion* into **build.gradle**, as simple variables. I don't ever need to override those on the command-line, so they don't really need to be Gradle *properties*.

russellsanborn:
Sounds good to me. I usually just put it in build.gradle too. I can update this PR for you or feel free to close it, Your call.

russellsanborn:
Updated. Like I said, feel free to close this if it is not what you want. And thanks again for CodeNarc. I've been enjoying digging in and understanding how things work more.

chrismair:
I will just make those values simple variables, I think, and only the ones used more than once. But thanks for the help.

PR #254, opened 2017-11-03
removed SuppressWarnings that were not needed.

russellsanborn:
In some cases the SuppressWarning was simply removed. For example, `EmptyWhileStatement` was being suppressed, but the while statement was not actually empty.

Other cases, the code needed small refactoring. For Example `NestedBlockDepth` warnings.


chrismair:
Thanks for the cleanup!

PR #255, opened 2017-11-05
Apply MissingOverrideAnnotation rule to project source code

erdi:
This PR also contains a fix to MissingOverrideAnnotationRule to not report violations for methods with default arguments for which not all possible signatures override methods. The defect was discovered after I applied the rule to project source code.


chrismair:
Holy crap! Wow, thanks for the fix and for that huge cleanup.

erdi:
It looks like a bigger task than it actually was. I managed to get the initial number of violation quickly down using a couple of regular expression replacements. :)

PR #258, opened 2017-11-05
Add enhanced mode for CloseWithoutCloseableRule.

erdi:
This PR also includes:
- a base class for dual mode rules called `AbstractEnhanceableAstVisitorRule`
- ability to enable enhanced mode for all dual mode rules using a system property
- simplification of testing of dual mode rules by introducing `GenericAbstractRuleTestCase`

This PR supersedes PR #27.


chrismair:
A few questions:
 1. I am not thrilled about introducing a new library dependency, especially for just one usage within a single test class. Are you okay with removing that?
 2. I see that introducing `GenericAbstractRuleTestCase` adds some type safety/documentation. But is it actually necessary anywhere?
 3. What is the rationale for having the *enhanced* test classes inherit from the regular test classes, e.g. `EnhancedCompareToWithoutComparableRuleTest`? Is it to ensure running all of the "regular" tests again in enhanced mode while also providing additional *enhanced*-only tests?

Thanks.

erdi:
Sorry for a delayed response, it's been a busy week.

1. While I agree that the dependency introduced with my change is not essential and I could reimplement clearing of the system property I change as part of the test I would like to understand your opposition against introducing a **test** dependency better. I personally do not see it as a cost to anybody (apart from people developing CodeNarc on their machines and the CI build who will have to download it once into their Gralde cache) because it doe not change anything to the users of CodeNarc and makes life easier to its developers. I also believe that the test reads better when using a rule instead of writing the clearing code in `@Before` and `@After` methods because it expresses the intent in a cleaner way.
As I said, I'm happy to amend the PR and remove the dependency but I would appreciate you reconsider given my arguments.

2. Introducing `GenericAbstractRuleTestCase` was indeed done for type safety. After it's introduction, the IDE understands that `rule` in `EnhancedCloseWithoutCloseableRuleTest` and `EnhancedCompareToWithoutComparableRuleTest` is an `AbstractEnhanceableAstVisitorRule` and provides autocompletion for its `enhancedMode` property in `@Before` methods. In my opinion it leads to cleaner code and describing your code to the IDE better enables more accurate refactorings in the future. In my view, I simply applied the Boy Scout Rule here.

3. Yes, _enhanced_ test classes inherit from the regular test classes for this specific reason. It allows me to ensure that the rule will yield the same results in both modes without having any test code duplication. While in the cases where enhanced rules so far no changes were needed for the enhanced mode to pass these tests you can potentially see that this will not always be the case, for example when class names are used in rule logic and they change when they are resolved in later compilation stages (i.e. simple name vs full class name, including the package). Do you have a suggestion for achieving the same test coverage in another way?

erdi:
I rebased onto latest master to resolve the conflicts and make it easier for your to merge it.

chrismair:
Okay, that all makes sense. Thanks for explaining.

Re: (1) above... I do not disagree. And perhaps we will find other uses for that library as well.

My hesitation is no doubt influenced by my own recent experience. I have spent the last year or so involved in a cleanup effort across many large Java/Groovy projects. Several of them had over many years accumulated various dependencies -- some of which were no longer needed, some were obsolete, and others were of questionable value. 

And those dependencies have an unfortunate habit of occasionally breaking or causing problems when you upgrade Java or Groovy or the dependency itself. Or in some cases, for us, use of the dependency was blacklisted by our centralized governance organization (Oh the joys of working in a large enterprise). And ripping them out or replacing them can be painful.

So in my experience, there is always a cost with adding stuff like that, but usually it is negligible, as I expect it will be in this case.

chrismair:
Excellent!. Thanks, Marcin.

erdi:
Thanks for sharing your view and experience with regards to working on cleaning up large projects. It helped me to put your hesitation to add any new dependencies into perspective. It also makes me happy that I managed to persuade you that pulling in that dependency was a good thing to do.

In my opinion software development is an exercise in risk mitigation. And quite often it's very hard to make the right choices for now let alone the future or the distant future. Will the lib I'm pulling in be maintained? Will it still work when I start updating things? Will it conflict with other deps I have?
 Maybe I should write it myself? But can I guarantee that the quality will be at the same level as of the library developed by people who are experts in the area? Won't it be too much work? And so on..

PR #259, opened 2017-11-05
Use earlier compiler phase in MissingOverrideAnnotation rule

erdi:
There is no need to use `CANONICALIZATION` in that rule, `SEMANTIC_ANALYSIS` will suffice, and using an earlier phase will save time by not having to perform another compilation phase on the analysed code.


chrismair:
Thanks Marcin!

Issue #236, opened 2017-09-23
UnnecessarySemicolon false positive on comments


grv87:
Apache 2.0 original license notice includes semicolon at the end of the line:
```groovy
/* ...
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
   You may obtain a copy of the License at
... */
```

CodeNarc tells that this is a violation of UnnecessarySemicolon.

Documentation says that `excludePattern` should exclude comments. Looks like its default value is not applied or is incomplete.

chrismair:
That rule uses regular expressions to find lines ending with a semicolon, and ignores lines matching several types of comments. So, if those lines had a leading asterisk, it would ignore them, e.g.

```
/*
 * Copyright 2011 the original author or authors.
 * 
 * Licensed under the Apache License, Version 2.0 (the "License");
 * you may not use this file except in compliance with the License.
```

I acknowledge that is just a workaround, and an annoying one at that.

grv87:
And if it is regular (non-comment) line with leading asterisk, is it ignored?

chrismair:
Yes.

grv87:
@chrismair, maybe CodeNarc should have a rule to check style of multiline comments (i.e. that they have leading asterisk). What do you think?

grv87:
Similar problem with `BlankLineBeforePackage` rule
False positive on blank lines in multi-line comments

chrismair:
> @chrismair, maybe CodeNarc should have a rule to check style of multiline comments (i.e. that they have leading asterisk). 

Yes, that sounds like it could be useful.

russellsanborn:
I submitted a PR that includes a new `MultilineCommentChecker` class that i think could be used not only for solving the issues mentioned here in `BlankLineBeforePackageRule` and `UnnecessarySemicolon`, but could be used for developing a multiline comment style rule.

Hope it helps

chrismair:
This was fixed in CodeNarc 1.1.

PR #260, opened 2017-11-10
Update to ignore semicolons in multiline comments

russellsanborn:
According to issue #236 multiline comments that do not have a leading asterisk were throwing a false positive. I updated the logic to contain a simple boolean that will keep track of whether the line is inside of a comment block or not.

### DIFF
--- a/src/main/groovy/org/codenarc/util/MultilineCommentChecker.groovy
+++ b/src/main/groovy/org/codenarc/util/MultilineCommentChecker.groovy
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.codenarc.util
+
+/**
+ * A class that is used to hold the state of a the presence of multiline comments.
+ *
+ * When processing files for violations, not all violations apply to comments. This class provides a simple way to
+ * process each line of code and determine if the current line exists inside of a multiline comment.
+ *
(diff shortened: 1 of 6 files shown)


russellsanborn:
@chrismair i don't know how you handle issues, so I decided not to use any of the github auto close issue hooks. Also, there was talk in #236 about fixing a similar problem with the `BlankLineBeforePackage` rule.

Let me know if there is anything more you would like me to do in terms of linking this to the issue.

russellsanborn:
Sorry about the multiple commits after I opened up the PR. I started working on the issue for `BlankLineBeforePackage` and realized they had the same solution so i refactored my original solution.

src/main/groovy/org/codenarc/util/MultilineCommentChecker.groovy
```diff
@@ -0,0 +1,55 @@
...
+ * process each line of code and determine if the current line exists inside of a multiline comment.
+ *
+ * @author Russell Sanborn
+ */
+class MultilineCommentChecker {
+    static final String START_MULTILINE_COMMENT = '/*'
+    static final String END_MULTILINE_COMMENT = '*/'
+    static final String START_MULTILINE_COMMENT_PATTERN = '.*/\\*.*'
+    static final String END_MULTILINE_COMMENT_PATTERN = '.*\\*/.*'
+
```
chrismair: If this is only used in tests, can you make it protected?

src/main/groovy/org/codenarc/util/MultilineCommentChecker.groovy
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * A class that is used to hold the state of a the presence of multiline comments.
+ *
+ * When processing files for violations, not all violations apply to comments. This class provides a simple way to
+ * process each line of code and determine if the current line exists inside of a multiline comment.
+ *
+ * @author Russell Sanborn
+ */
+class MultilineCommentChecker {
```
chrismair: Can you make these private?

src/main/groovy/org/codenarc/util/MultilineCommentChecker.groovy
russellsanborn: updated

src/main/groovy/org/codenarc/util/MultilineCommentChecker.groovy
russellsanborn: updated

chrismair:
Thanks!

PR #262, opened 2017-11-10
Update readme to describe how to test locally

russellsanborn:
I made the suggested command match the command in `.travis.yml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,11 @@ AVAILABLE FROM MAVEN CENTRAL REPOSITORY
 For projects built using Gradle or Maven, **CodeNarc** is available from the Maven Central Repository.
   - groupId = org.codenarc
   - artifactId = CodeNarc
+
+## Development
+
+When contributing to CodeNarc, run the following command to test on your local machine.
+
+```bash
+./gradlew clean check
+```


russellsanborn:
If you would rather have something like this in a `CONTRIBUTING.md`, let me know.

chrismair:
Thanks for the contribution!

README.md
```diff
@@ -18,3 +18,11 @@ AVAILABLE FROM MAVEN CENTRAL REPOSITORY
 For projects built using Gradle or Maven, **CodeNarc** is available from the Maven Central Repository.
   - groupId = org.codenarc
   - artifactId = CodeNarc
+
+## Development
+
+When contributing to CodeNarc, run the following command to test on your local machine.
+
+```bash
+./gradlew clean check
```
erdi: @chrismair Asking people to always run `clean` task is bad practice because you loose all the benefits of [Gradle's incremental build features](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks) by doing so.

russellsanborn: Good call out. I added `clean` because I always run it like that before submitting my PR, but what you are saying makes total sense.

PR #263, opened 2017-11-10
Fix false positive when an annotation has a brace

Brickster:
If a constructor contained a brace, `SpaceAfterOpeningBraceRule` would report a false positive:

```
class MyClass {
  MyClass(@Value('${property}') String property) {

  }
}
```

It is not uncommon for Spring projects that use this mechanism to load properties into the construction of an object.


chrismair:
Thanks for the contribution!

PR #265, opened 2017-11-11
Remove "clean" task from suggested gradle test command

russellsanborn:
Thanks to @erdi for suggesting this be removed based on the link they provided:
https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks


chrismair:
Thanks.

PR #286, opened 2017-12-28
Add MethodReturnTypeRequired rule.

### LINKED ISSUE #285: New MethodReturnTypeRequired rule
Add a rule which reports violations for methods which use `def` as a return type or do not specify return type at all.

erdi:
Closes #285.

Currently the rule is not applied against project source code as that causes almost 200 violations. Please let me know if you would like it to be applied and the violations fixed.


chrismair:
Looks good. Thanks.

If we want to clean up the violations in the CodeNarc code, let's tackle that as a separate PR. I consider that optional. Or maybe just clean up the violations in the main source and ignore tests.

PR #287, opened 2017-12-28
Fix encoding of changelog file 

erdi:
The changelog contains accents in people's names which got garbled over time and get even more garbled every time one uses the create-task script - this PR alsp updates the script to address that in a cross platform way.


chrismair:
Nice. Thanks.

PR #289, opened 2017-12-28
Add MethodParameterTypeRequired rule.

### LINKED ISSUE #288: New MethodParameterTypeRequired rule
Add a rule which reports violations for methods which use `def` as a parameter type or do not specify a parameter type at all.

erdi:
Closes #288.

Currently the rule is not applied against project source code as that causes almost 150 violations. Please let me know if you would like it to be applied and the violations fixed.


chrismair:
Thanks Marcin!

Let's handle any cleanup (optional) as a separate PR.

PR #292, opened 2017-12-29
Add BlockStartsWithBlankLine rule

### LINKED ISSUE #291: New BlockStartsWithEmptyLine formatting rule
Add a rule which reports violations if a code block (method body, closure, if/else bodies, loop body) starts with an empty line.

erdi:
Closes #291.


chrismair:
Can you add some more tests for Closures with different formatting options, including:
 * Just brace (no arrow) on same line as declaration
 * Brace and arrow on same line as declaration
 * Just brace (no arrow) on  next line
 * Brace on same line, arrow on next line
 * Multiple params on multiple lines

e.g.
```
list.each {
  // blank
}

list.each { item ->
  // blank
}

list.each { item     // is this valid; and is it relevant?
    ->
  // blank
}

list.each { 
    item ->
  // blank
}

map.each { 
    def key,
    def value ->
  // blank
}
```

erdi:
Of course I can. Please be aware that it probably won’t be instant, though.

erdi:
I've added a test for violations caused by the first two cases you've given. The other three are not causing violations but given that they are fairly edge case I decided to just add a test documenting them as known limitations. I can update documentation for the rule to mention them explicitly if you think it's worth it.

chrismair:
Looks Good. Thanks Marcin!

Issue #244, opened 2017-10-18
JUnit* rules are not compatible with JUnit5


ychulovskyy:
JUnit rules shows a lot of false positives for JUnit 5 because annotation names are different.
1. Rule=JUnitPublicNonTestMethod has false positive "Violation in class <class name>. The method setup is public but not a test method" because there is @BeforeEach annotation above a method instead of @Before
2. Rule=JUnitSetUpCallsSuper has false positive "Violation in class <class name>. The method setUp() does not call super.setUp()" because there is @BeforeEach annotation above a method instead of @Before

Those are just examples.
We can use http://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4-tips as reference.
I can work on this if you approve the change.

chrismair:
Sounds good, assuming you can keep those rules backward-compatible with JUnit 4. Thanks.

chrismair:
I may start working on this unless someone else is intending to...

ychulovskyy:
@chrismair I plan to work o it next week, if you don't mind

chrismair:
Excellent. Thanks.

ychulovskyy:
Closing as resolved

PR #295, opened 2017-12-29
#244 JUnit* rules are not compatible with JUnit5

ychulovskyy:


src/main/groovy/org/codenarc/rule/junit/JUnitPublicNonTestMethodRule.groovy
```diff
@@ -50,6 +50,13 @@ class JUnitPublicNonTestMethodRule extends AbstractAstVisitorRule {
     int priority = 2
     Class astVisitorClass = JUnitPublicNonTestMethodAstVisitor
     String applyToClassNames = DEFAULT_TEST_CLASS_NAMES
```
chrismair: The fact that *ignoreMethodsWithAnnotations* is a property means it is publicly accessible. It is confusing to have a setter with the same name but different type. Did you intend this to be *overloaded*? It seems like either *ignoreMethodsWithAnnotations* should be a private field or else the property should have a different name.

src/site/apt/codenarc-rules-junit.apt
```diff
@@ -208,9 +208,11 @@ JUnit Rules  ("<rulesets/junit.xml>")
...
 
-      * Methods annotated with <<<@Before>>> and <<<@After>>>
+      * Methods annotated with <<<@Before>>>, <<<@BeforeAll>>>, <<<@BeforeClass>>> and <<<@BeforeEach>>>
 
-      * Methods annotated with <<<@BeforeClass>>> and <<<@AfterClass>>>
+      * Methods annotated with <<<@After>>>, <<<@AfterAll>>>, <<<@AfterClass>>> and <<<@AfterEach>>>
+
+      * Methods annotated with <<<@Disabled>>> and <<<@Ignore>>>
 
       * Methods annotated with <<<@Override>>>
```
chrismair: Can you add a table row for the rule configuration property (*ignoreMethodsWithAnnotations*).

src/main/groovy/org/codenarc/rule/junit/JUnitPublicNonTestMethodRule.groovy
ychulovskyy: Good catch! Fixed

src/site/apt/codenarc-rules-junit.apt
ychulovskyy: Fixed

ychulovskyy:
@chrismair Happy New Year!

chrismair:
Thanks for the contribution!

PR #296, opened 2017-12-31
Enable MethodParameterTypeRequired and MethodReturnTypeRequired rules…

erdi:
… when analyzing project source code.


src/test/groovy/org/codenarc/ant/CodeNarcTaskAccessor.groovy
```diff
@@ -1,50 +0,0 @@
...
-package org.codenarc.ant
-
-import org.apache.tools.ant.types.FileSet
-
-/**
- * Test-only class for accessing protected members of CodeNarcTask from test classes in other packages
- *
- * @author Chris Mair
- */
-class CodeNarcTaskAccessor {
```
erdi: This class had violations in it but also was not used anywhere so I decided that it's better to remove it completely.

chrismair:
Whew! Thanks a lot for that big cleanup!

Issue #280, opened 2017-12-26
Use generics (GenericAbstractRuleTestCase) in all Rule tests


chrismair:
Use generics (`GenericAbstractRuleTestCase`) in all Rule tests for better type-safety, self-documentation and IDE completion.

Perhaps collapse `GenericAbstractRuleTestCase` back into `AbstractRuleTestCase`.

erdi:
FYI, I'm picking this up.

chrismair:
Excellent. Thanks.

chrismair:
Oh, please include updating **templates/Test.groovy** as appropiate.

chrismair:
Available with CodeNarc 1.1

PR #297, opened 2017-12-31
Use GenericAbstractRuleTestCase as a base for all rule tests cases

erdi:
Part of #280.

I refrained from collapsing `GenericAbstractRuleTestCase` into `AbstractRuleTestCase` because it's a breaking change for anyone using `AbstractRuleTestCase` to test their custom rules. I'll leave the decision to you but if you want to go ahead then you can do it by:
- deleting `AbstractRuleTestCase`
- renaming `GenericAbstractRuleTestCase` to `AbstractRuleTestCase`
- updating `templates/Test.groovy`

### DIFF
--- a/src/test/groovy/org/codenarc/rule/AbstractAstVisitorRuleTest.groovy
+++ b/src/test/groovy/org/codenarc/rule/AbstractAstVisitorRuleTest.groovy
@@ -15,12 +15,12 @@
  */
 package org.codenarc.rule
 
+import static org.codenarc.test.TestUtil.shouldFailWithMessageContaining
+import static org.junit.Assert.assertFalse
+
 import org.codehaus.groovy.ast.ClassNode
 import org.codenarc.util.WildcardPattern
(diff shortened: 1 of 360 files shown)


chrismair:
I am a bit confused. Last week I made a temporary local change to add generics to `AbstractRuleTestCase`. I changed a couple tests to specify generic types but left the rest as is, and everything ran fine. Did I miss something; what would break?

erdi:
Right, I forget that Groovy is very lenient when it comes to generics and you can quite often just omit them. So given a generic type of `T extends Rule` for `AbstractRuleTestCase` the following class would still compile and run:
```
class SomeRuleTest extends AbstractRuleTestCase {
    Rule createRule() {
         new SomeRule()
    }
}
```

This means that any classes extending from `AbstractRuleTestCase` _probably_ will not need to change after introducing generics to it.

chrismair:
Thanks a lot for that cleanup work!

PR #315, opened 2018-01-26
Fix Indentation Rule to work with spock block labels

### LINKED ISSUE #308: Indentation Rule reports errors with labels
This is particularly problematic when used with spock.

```groovy
class MySpec extends Specification {
    void 'some feature'() {
        given: 'something'
        def a = new Object()

        when: 'something is done'
        def b = a.toString()

        then: 'something happens'
        b != ''
    }
}
```

Codenarc reports errors with Indentation on the lines with labels. It appears codenarc thinks of the labels as whitespace.

russellsanborn:
I know there has been discussion on here before that spock can make detecting issues more complicated. I think the logic in `isSpockBlock()` should give good enough coverage. Also, since it is a skip, it will not falsely mark any lines as violations.

This fixes #308


chrismair:
Thanks for the contribution.

PR #318, opened 2018-02-09
Allow newline after elvis operator

russellsanborn:
If there was a newline after an elvis operator, it was throwing a SpaceAfterOperator warning.

Previously the code was set up to make sure at least one elvis operator had a space after it. Unfortunately that meant the following line would have passed:

```groovy
def doubleElvis = fullname ?:lastname ?: 'me'
```

I updated the check if there is instance of an elvis operator with any non-whitespace character following it.

If you like this method, it could be explored if it is causing problems with any other operators too.

### DIFF
--- a/src/test/groovy/org/codenarc/rule/formatting/SpaceAroundOperatorRuleTest.groovy
+++ b/src/test/groovy/org/codenarc/rule/formatting/SpaceAroundOperatorRuleTest.groovy
@@ -212,12 +212,11 @@ class SpaceAroundOperatorRuleTest extends AbstractRuleTestCase<SpaceAroundOperat
                     def greeting = fullname?:'you'
                     def f = funds.collect {it.fundSortOrder}?:[]
                     assert model.UserID == expectedModel.UserID?:null
-                    def doubleElvis = fullname ?:lastname ?: 'me'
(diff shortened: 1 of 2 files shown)


src/test/groovy/org/codenarc/rule/formatting/SpaceAroundOperatorRuleTest.groovy
```diff
@@ -212,16 +212,36 @@ class SpaceAroundOperatorRuleTest extends AbstractRuleTestCase<SpaceAroundOperat
                     def greeting = fullname?:'you'
                     def f = funds.collect {it.fundSortOrder}?:[]
                     assert model.UserID == expectedModel.UserID?:null
+                    def doubleElvis = fullname ?:lastname ?: 'me'
                 }
             }
         '''
+
+        // Line 7 has two violations because it evaluates the line twice, once for each elvis operator
```
chrismair: But it does not seem valid that there are two violations for that line; only one operator is not followed by whitespace.

russellsanborn:
@chrismair I updated this to use the`expression.columnNumber` instead after thinking about it more. It now correctly throws only the accurate number of warnings and I updated with a few more test cases.

chrismair:
Cool. Thanks for the contribution!

PR #323, opened 2018-03-02
Add terminating semicolons to keep consistency in js sort functions

russellsanborn:


chrismair:
Thanks for the cleanup!

PR #324, opened 2018-03-02
Add missing Violation type to List parameter in applyTo methods

russellsanborn:


chrismair:
Thanks for the cleanup!

PR #325, opened 2018-03-02
Perform general cleanup across project

russellsanborn:
This included a few different things:
* removing not needed return statement
* making a few call chains null safe for the entire chain
* simplifying a double negative boolean expression to be positive


chrismair:
The unnecessary `return` and fixing the double negation look good.

But re: adding more null-safe operators: I am hesitant to add more conditionals (branches) without accompanying tests. Are these actually fixes for NPEs you encountered?

russellsanborn:
I didn't encounter and NPEs. I was just trying to keep the call chain null safe. 

I can remove those pieces for now, dig in a little more to see the best way to write tests for it (this is a new section in the project I would want to spend a little time in first :smile: ), and make a new PR in the future.

chrismair:
Cool. Thanks.

PR #351, opened 2018-07-12
Add `grailsApplication` to the default ignored fields in a stateless service

doelleri:
In Grails 3.x it is increasingly common to statically type the injected grailsApplication as `GrailsApplication grailsApplication`. This adds grailsApplication to the default ignored fields for a stateless service.


chrismair:
Thanks for the contribution

PR #354, opened 2018-07-26
Ignore ClassRule in JUnitPublicFieldRule

daspilker:
`JUnitPublicFieldRule` already ignores the `@Rule` annotation and should also ignore `@ClassRule`.


chrismair:
Thanks for the contribution

PR #362, opened 2018-09-10
Add ClassStartsWithBlankLine rule

tarkil:
Check whether the class starts with a blank line. By default, it enforces that there must be a blank line after the opening class brace, except if the class is empty and is written in a single line. A blank line is defined as any line that does not contain any visible characters.


benatespina:
It looks great!!

PR #365, opened 2018-09-28
Reset switch counter on method visit

### LINKED ISSUE #363: CouldBeSwitchStatement does not consider scope
I ran into a violation of `CouldBeSwitchStatement` in unit test code which looked a bit like this:

```groovy
@Test
void testA() {
  def closure = { i ->
    if (i == 'foo') {
      throw new Exception()
    }
  }
  // etc...
}

@Test
void testB() {
  def closure = { i ->
    if (i == 'bar') {
      throw new Exception()
    }
  }
  // etc...
}

@Test
void testC() {
  def closure = { i ->
    if (i == 'baz') {
      throw new Exception()
    }
  }
  // etc...
}
```

Changing the variable name fixes the violation, which is a hacky workaround for the time being.

jlstrater:
resolves #363


chrismair:
Looks good. Thanks for the fix!

PR #366, opened 2018-09-28
add option to ignore regular imports from wildcard imports rule

### LINKED ISSUE #360: NoWildcardImports: support for only static violation (suggestion)
I think it would be sensible to add support for only static violation in [NoWildcardImports](http://codenarc.sourceforge.net/codenarc-rules-imports.html#NoWildcardImports).

Ie. currently `ignoreStaticImports` is supported, I'm suggesting `ignoreImports` with default `false`.

That way it's possible to allow wildcard import for `class` references, and simultaneously mark `static` wildcard imports as violations.
As the implications for `static` wildcard imports can be far more severe and subtle as everything (methods, fields, etc.) in `static` context is "merged".

jlstrater:
#360 seems like a reasonable thing to ask for so I added the config option.

resolves #360


chrismair:
Thanks for the contribution!

PR #367, opened 2018-09-28
ignore violation when private field is annotated with @Lazy

### LINKED ISSUE #357: PrivateFieldCouldBeFinal for @Lazy field
`PrivateFieldCouldBeFinal` should not be reported if the field is annotated with `@Lazy` since such field cannot be final.

jlstrater:
resolves #357


chrismair:
Looks good. Thanks for the fix!

PR #372, opened 2018-10-17
Also handle the "Spy" Spock method in UnnecessaryGetterRule

boris-petrov:


chrismair:
Thanks for the contribution!

PR #381, opened 2018-12-06
Update CloseWithoutCloseableRule to not create a violation for classes implementing Autocloseable

msanssouci:


chrismair:
Thanks for the contribution!

PR #391, opened 2019-01-15
Fix codenarc.groovy

darxriggs:
Running the script via `./codenarc.groovy` which uses the interpreter provided via the shebang only works if Linux line endings "\n" are used.


chrismair:
Thanks for the fix!

PR #392, opened 2019-01-15
Fix compilation of Groovy class in incorrect folder

darxriggs:
For Gradle this wasn't a problem but compilation failed under Eclipse.
Therefore convert the only Groovy class in src/main/java to a Java class.


chrismair:
This changes *enhancedMode* from a property to a public field. I'd be inclined to keep it as a property: private field + getter/setter.

darxriggs:
I changed it to a property.

chrismair:
Great. Thanks for that.

PR #393, opened 2019-01-15
Remove obsolete Java placeholder file

darxriggs:
It was only required by the Groovy-Eclipse compiler.
The project was migrated from Maven to Gradle in 2015.
Therefore this compiler is not used anymore.

The initial cause why this file was required can be found under:
https://github.com/groovy/groovy-eclipse/wiki/Groovy-Eclipse-Maven-plugin#do-almost-nothing


chrismair:
Thanks for the clean up!

PR #406, opened 2019-03-02
Add ignoreMethodNames property to UnnecessaryGetter 

### LINKED ISSUE #405: UnnecessaryGetter: Allow ability to ignore certain methods
Consider this code 

```groovy

Optional<string> foo

foo.getOrNull()
```

The `UnnecessaryGetter` rule will complain that it should be written as `orNull`. Although this might be a case of personal preference for people, `getOrNull` is not a true getter. For cases where people will prefer to keep the getXYZ method call, instead of the Groovy-style property, it will be worth for the rule to have a configurable way to ignore certain method names.

erdi:
Fixes #405.


chrismair:
Cool. Thanks for the contribution!

PR #412, opened 2019-03-14
IndentationRule: Add test on an array containing a closure param

debanne:
This commit is just here to illustrate a bug


chrismair:
Thanks for the heads up and the pre-made test.

PR #418, opened 2019-05-21
Add new convention rule for CompileStatic

snimavat:
Add new CompileStatic rule which can be used to ensure that classes are annotated with compile static. and if the class does not need to be compile static, it has to be explicitely annotated with @CompileDynamic

### DIFF
--- a/src/main/groovy/org/codenarc/rule/convention/CompileStaticRule.groovy
+++ b/src/main/groovy/org/codenarc/rule/convention/CompileStaticRule.groovy
@@ -1,5 +1,5 @@
 /*
- * Copyright 2009 the original author or authors.
+ * Copyright 2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,7 +15,6 @@
  */
 package org.codenarc.rule.convention
 
-import groovy.transform.CompileStatic
 import org.codehaus.groovy.ast.AnnotationNode
 import org.codehaus.groovy.ast.ClassNode
 import org.codehaus.groovy.ast.InnerClassNode
@@ -28,14 +27,15 @@ import org.codenarc.rule.AbstractAstVisitorRule
  *
  * @Author Sudhir Nimavat
  */
-@CompileStatic
 class CompileStaticRule extends AbstractAstVisitorRule {
+
     int priority = 2
     String name = 'CompileStatic'
     Class astVisitorClass = CompileStaticlVisitor
 }
 
 class CompileStaticlVisitor extends AbstractAstVisitor {
+
     private static final String GRAILS_COMPILE_STATIC = 'GrailsCompileStatic'
     private static final String COMPILE_STATIC = 'CompileStatic'
     private static final String COMPILE_DYNAMIC = 'CompileDynamic'
--- a/src/main/groovy/org/codenarc/rule/convention/CompileStaticRule.groovy
+++ b/src/main/groovy/org/codenarc/rule/convention/CompileStaticRule.groovy
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2009 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


snimavat:
Please review and comment if it needs any changes.

jlstrater:
what about the case where static compilation is added via Gradle?

chrismair:
I am also interested in the answer to the question from @jlstrater re: what if @CompileStatic is configured through Gradle, across the project. I would assume you then not use/enable this rule, right?

src/main/groovy/org/codenarc/rule/convention/CompileStaticRule.groovy
```diff
@@ -0,0 +1,65 @@
+/*
```
chrismair: Please update to 2019

src/main/groovy/org/codenarc/rule/convention/CompileStaticRule.groovy
```diff
@@ -0,0 +1,65 @@
...
+import org.codenarc.rule.AbstractAstVisitor
+import org.codenarc.rule.AbstractAstVisitorRule
+
+/**
+ * Enforces classes are annotated either with one of the CompileStatic, GrailsCompileStatic or CompileDynamic
+ * annotations
+ *
+ * @Author Sudhir Nimavat
+ */
+@CompileStatic
```
chrismair: Why is @CompileStatic necessary/useful here? It is certainly not conventional in this code base.

src/main/groovy/org/codenarc/rule/convention/CompileStaticRule.groovy
```diff
@@ -0,0 +1,65 @@
...
+class CompileStaticRule extends AbstractAstVisitorRule {
+    int priority = 2
```
chrismair: Please add a blank line at the top of each class.

src/test/groovy/org/codenarc/rule/convention/CompileStaticRuleTest.groovy
```diff
@@ -0,0 +1,94 @@
+/*
```
chrismair: Please update to 2019

src/test/groovy/org/codenarc/rule/convention/CompileStaticRuleTest.groovy
```diff
@@ -0,0 +1,94 @@
...
+        final SOURCE = '''
+            import groovy.transform.CompileDynamic
+            @CompileDynamic
+            class Test {
+                class AnInnerClass { }
+            }
+         '''
+        assertNoViolations(SOURCE)
+    }
+}
```
chrismair: Add test for Enum.
Add test for abstract class.

src/test/groovy/org/codenarc/rule/convention/CompileStaticRuleTest.groovy
```diff
@@ -0,0 +1,94 @@
...
+import org.codenarc.rule.Rule
+import org.codenarc.rule.Violation
+import org.junit.Test
+
+/**
+ * Tests for CompileStatic rule.
+ *
+ * @author Sudhir Nimavat
+ */
+class CompileStaticRuleTest extends AbstractRuleTestCase {
```
chrismair: Add blank line at top of class

basejump:
I have been working with @snimavat on this and can answer the question of static compilation being added via Gradle.

1. yes, I would assume that if you were doing gradle compile static that then you would not enable this
2. we had problems with getting Idea IDE editor to pick up the gradle setting where it does read the annotation
3. might be angry monkey memory but I recall the global compileStatic being finiky about overriding it with @CompileDynamic and speicifically trying to override it when usign Grails @GrailsCompileStatic
4. we wanted it to be explicit what the intent is on each class so it was immediately obvious if it was @CompileStatic, @CompileDynamic or @GrailsCompileStatic

snimavat:
@chrismair I have made changes, pl review.

chrismair:
Looks good. Thanks for the updates, and thanks for the contribution.

PR #421, opened 2019-05-27
Update to gradle 5.4.1, Only sign on uploadArchives

leonard84:


chrismair:
Thanks for the contribution. This should probably have been 2 separate PRs. Is the "Only sign on uploadArchives" part necessary, or just an optimization? Was that causing problems?

leonard84:
@chrismair if you don't have any credentials set, then a simple `gradlew build` will fail, because sign complains that it does not have the necessary info.

chrismair:
Thanks for the contribution!

PR #423, opened 2019-05-27
Update numbering scheme for groovy 3

### LINKED ISSUE #414: Better support for the Parrot parser
I found when using CodeNarc within the Groovy 3 build that some files were failing CodeNarc processing. There are very minor changes in how the new parser records line and column information to improve accuracy of that information and those changes seem to trip up processing in a few places. I'll do a bit more testing and then create a PR but the following diff seemed to fix most issues if anyone else wants to test in parallel. This seemed to fix most issues using Groovy 3 and didn't see to create new ones with 2.5. I did notice that I was getting worse false positives for rules being run on text within triple quoted strings. E.g. for a template generating Java source code it was telling me that the semicolons weren't needed. I'll investigate some more. The -2 to -1 change in the diff is perhaps coming into play there.

```
Index: src/main/groovy/org/codenarc/rule/unnecessary/UnnecessaryPublicModifierRule.groovy
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/main/groovy/org/codenarc/rule/unnecessary/UnnecessaryPublicModifierRule.groovy	(revision 9a7cec02cb8cbaf845030f2434309e8968f5d7a7)
+++ src/main/groovy/org/codenarc/rule/unnecessary/UnnecessaryPublicModifierRule.groovy	(date 1554416536324)
@@ -80,6 +80,9 @@
         while (current <= node.lastLineNumber) {
             def line = sourceCode.line(current)
             if (line == null) {
+                if (node.isAbstract()) {
+                    return acc // can happen with abstract method in Parrot parser
+                }
                 LOG.warn("${rule.name} cannot find source code line $current in ${sourceCode.name}. Scanning lines ${node.lineNumber} to ${node.lastLineNumber}.")
                 return ''
             } else if (line.contains('{')) {
Index: src/main/java/org/codenarc/util/AstUtil.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- src/main/java/org/codenarc/util/AstUtil.java	(revision 9a7cec02cb8cbaf845030f2434309e8968f5d7a7)
+++ src/main/java/org/codenarc/util/AstUtil.java	(date 1554377573185)
@@ -1197,7 +1197,7 @@
 
             // extract the relevant part of the last line
             if (lineIndex == node.getLastLineNumber() - 1) {
-                int stopIndex = node.getLastColumnNumber() < line.length() ? node.getLastColumnNumber() - 2 : line.length();
+                int stopIndex = node.getLastColumnNumber() < line.length() ? node.getLastColumnNumber() - 1 : line.length();
                 line = line.substring(0, stopIndex);
             }
 
```

philippkrauss:
We started this pull request on Gr8conf 2019. It actually fixes the issues described in issue #414.
It still needs to be tested further.
resolves #414

### DIFF
--- a/src/main/groovy/org/codenarc/rule/unnecessary/UnnecessaryPublicModifierRule.groovy
+++ b/src/main/groovy/org/codenarc/rule/unnecessary/UnnecessaryPublicModifierRule.groovy
@@ -80,6 +80,9 @@ class UnnecessaryPublicModifierAstVisitor extends AbstractAstVisitor {
         while (current <= node.lastLineNumber) {
             def line = sourceCode.line(current)
             if (line == null) {
+                if (node.isAbstract()) {
+                    return acc // can happen with abstract method in Parrot parser
+                }
                 LOG.warn("${rule.name} cannot find source code line $current in ${sourceCode.name}. Scanning lines ${node.lineNumber} to ${node.lastLineNumber}.")
                 return ''
             } else if (line.contains('{')) {
--- a/src/main/java/org/codenarc/util/AstUtil.java
+++ b/src/main/java/org/codenarc/util/AstUtil.java
@@ -1197,7 +1197,7 @@ public class AstUtil {
 
             // extract the relevant part of the last line
             if (lineIndex == node.getLastLineNumber() - 1) {
-                int stopIndex = node.getLastColumnNumber() < line.length() ? node.getLastColumnNumber() - 2 : line.length();
+                int stopIndex = node.getLastColumnNumber() < line.length() ? node.getLastColumnNumber() - 1 : line.length();
                 line = line.substring(0, stopIndex);
             }
 


chrismair:
What is the most straightforward way to see that this solves the Groovy 3 problems? If we (temporarily) switch the CodeNarc project to use Groovy 3 (beta), will the existing test suite fail? If not, can we add some test cases that do fail under that, even if they pass with Groovy 2.x? I am hesitant to pull in code changes that don't have any test coverage.

jlstrater:
good question. The problem with switching the project itself is that it doesn't compile under Groovy 3. I could try to come up with something that works for the tests though.

chrismair:
I was able to reproduce the problems by running CodeNarc locally against the Groovy 3.0.0-beta-2 source code. I merged the changes above manually because of the AstUtil conflict.

Fixes #414 

Thanks for the contribution.

PR #424, opened 2019-05-29
Add ImplicitClosureParameter rule.

### LINKED ISSUE #371: Rule for use of implicit closure parameters
Some developers want to avoid magic like the `it` variable, to reduce risk. Could we have a CodeNarc rule that encourages coders to use explicit closure variables?

http://groovy-lang.org/closures.html

erdi:
Fixes #371.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CodeNarc Change Log
 
+Version 1.5
+--------------------------------------
+New Rules
+- #371: **ImplicitClosureParameter** rule (convention). Checks for usages of the implicit `it` closure parameter.
+
 Version 1.4   (May 2019)
 --------------------------------------
 New Rules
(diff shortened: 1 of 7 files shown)


chrismair:
Thanks for the contribution, Marcin.
1. Are the huge diffs with existing files because of line-endings? Can we avoid/revert that?
2. Add a test for no-arg closure with code, e.g. `{ println 123 }`.

erdi:
Oh sorry, I wasn't aware that the line endings were changed. I will revert, of course. I will also add the test you asked for.

erdi:
All that you've asked for should be there now, @chrismair.

chrismair:
I have to admit, I would expect the behavior of this rule to be that `{ println 123 }` would not produce a violation, since it does not use `it`. That's what motivated me asking for that test to be added. Are you thinking there are teams out there that insist that `->` always be specified within all closures?

erdi:
Yeah, you seem to have a point there. Pretty pointless to have a violation when you are actually not referencing `it`. And thinking that I only decided to submit a PR for this feature request cause it seemed like an easy one to implement. 🤣

erdi:
So, how is it looking now?

erdi:
So, how is it looking now?

chrismair:
Nice. Thanks Marcin.

PR #429, opened 2019-06-14
Normalize line endings to LF.

erdi:
This PR is an effort to solve my line ending issues which arose when I submitted https://github.com/CodeNarc/CodeNarc/pull/424.

I believe that industry standard when project files are modified on OSes which use both CRLF (Windows) and LF (Unix-like) is to use LF as the endings in git so that line endings are not mixed which should make diffs and merging easier.

I've added a `.gitattributes` file to always normalise line endings to LF regardless of the global `core.autocrlf` setting and keep CRLF as line endings for `*.bat` files which are Windows only.

The diff is huge and it might be daunting at first, but if you review it in IntelliJ you will easily see that all that's changed are the line endings.

I've followed this guide while working on this, if you're interested in rationale: https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/#the-new-system and used this to normalise line endings across the board: https://help.github.com/en/articles/dealing-with-line-endings#refreshing-a-repository-after-changing-line-endings.

If you believe that we should not be normalising line endings (and currently there are files with both CRLF and LF line endings in the repo) then an alternative would be to add a `.gitattributes` file with the following contents:
```
* -text
```
which will consistently prevent any normalisation regardless of what the global setting for `core.autocrlf` is on a given machine.


chrismair:
I had to merge manually, fix the conflicts, and fix one failing test in `SourceFileTest` that was based on character indexes within a text file. I think/hope I did all that correctly.

erdi:
Great, thanks for applying this even though it was not straightforward.

PR #431, opened 2019-07-10
Update doc

TobiX:
Since I just released a new SonarQube Groovy plugin, this fixes the links to point to my new version.

While I was at it, I replaced Hudson with Jenkins and recommend an active plugin for that. I didn't touch the  `InlineXmlReportWriter` since that looks specific to the old Violations plugin... (Maybe it can be deprecated/removed?)


chrismair:
Thanks for the updates!

PR #462, opened 2020-03-02
ClassStartsWithBlankLine rule now supports annotations with closures

tlefevre:
The rule should now be able to handle annotations that contain a closure. 

I ran into this problem when using the Spock Requires annotations which takes a closure. I have tried to think of all the possible ways to combine annotations and such and implement test cases for them.


tlefevre:
Would those not be testNoViolationsWithSingleClassWithSimpleAnnotation and testNoViolationsWithSingleClassWhenAnnotationContainsOpeningBrace ?

chrismair:
Thanks for the contribution!

PR #464, opened 2020-03-04
MethodReturnTypeRequiredRule now has an ignoredMethods property

tlefevre:
http://gorm.grails.org/6.1.x/hibernate/manual/#eventsAutoTimestamping

Ran into a problem where these methods were being flagged. I would rather just that Codenarc ignore these methods.


chrismair:
To be consistent with the common pattern/convention used in other rules, please name the new property `ignoreMethodNames` and access it as `rule.ignoreMethodNames` from the AstVisitor. See `MethodNameRule` and `IllegalClassMemberRule` as examples.

chrismair:
Thanks for the contribution.

PR #466, opened 2020-03-04
MethodParameterTypeRequiredRule now has an ignoredParameters property

tlefevre:
When working with parsed json in Groovy, we use "def json". It would be nice to add this as an ignored parameter type.

JsonSlurpers parse methods and other json tools in Groovy generally use def as the type, with the interface/usage generally being the same between implementations. Having to use the specific implementation types would not be particularly helpful for our developers.


chrismair:
To be consistent with the common pattern/convention used in other rules, please name the new property `ignoreMethodNames` and access it as `rule.ignoreMethodNames` from the AstVisitor. See `MethodNameRule` and `IllegalClassMemberRule` as examples.

chrismair:
Thanks for another contribution!

PR #468, opened 2020-03-12
StatelessClassRule now ignores fields annotated with @Value

tlefevre:
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/annotation/Value.html

Getting some violations for Value annotated fields, but they are dependency injection fields handled by Spring.


chrismair:
Nice. Thanks.

PR #470, opened 2020-03-26
OpenJDK 11 and 14 builds on travis + required changes

szpak:
This PR adds Openjdk 11 and 14 builds on Travis. Changing target compatibility to (at least) 1.7 is required to make the binary runnable with Java 14 (see https://github.com/gradle/gradle/issues/12646).

As there are some problems reported in #471, I keep it as draft to make decisions how to handle them.

Btw, Travis doesn't seem to be executed for draft PRs, to [here](https://travis-ci.org/github/szpak/CodeNarc/builds/667344019) are Travis builds for it.

Btw2, having a need to bump source compatibility to 1.7 for Java 14 builds (and maybe also runtime compatibility?), maybe it would be good to switch straight to 1.8 (and maybe also upgrade Groovy to 2.5)?


szpak:
I rebased my PR after your changes in #471.

Source compatibility with 1.6 could be preserved with some extra scripting in Travis, however, most of the world usually use Java 1.8 anyway.

szpak:
Btw, haven't you enabled Travis builds for PRs?

chrismair:
>> Btw, haven't you enabled Travis builds for PRs?

Travis: *Build pushed pull requests* is already set to *true/on*.

I'm not inclined to bump *sourceCompatibility* for the next (minor) release, since that seems a clear breaking change for users still stuck on old Java, if there are any. I am fine doing that when we go to 2.0.

chrismair:
#519 upgraded Gradle Wrapper to 6.5. Java 14 support will be handled in #525.

PR #476, opened 2020-04-01
Fix issue with missing escape for pipe

Mazorius:
Inside section Rule: Size a pipe inside a table was not escaped which results in bad output.


chrismair:
Cool. Thanks for the fix!

Issue #477, opened 2020-04-02
Return CodeNarc version via command line ?


nvuillam:
I looked in the repo, but I didn't see :
- **1**. option **-v** / **-version**
- **2**. any way to find in the source the version of codenarc

If there is **2**. , i could make a PR to manage **1**.
If not, could it be possible to plan **1 & 2** in a future version ?

Many thanks !

chrismair:
Great idea. There is logic in `AbstractReportWriter`, but I am inclined to move that out into its own `CodeNarcVersion` class. I can do that part.

chrismair:
Okay, `CodeNarcVersion` is available now.

nvuillam:
Great !
I created #478 so it can be returned via command line :)
When do you plan to release the next CodeNarc version ?

chrismair:
Next version of CodeNarc is likely in May or June.

chrismair:
Fixed by #478.

PR #478, opened 2020-04-03
Manage -version argument when calling via cmd/bash +groovylintrc.json

nvuillam:
- Add -version argument management + test method
- Apply CodeNarc rules on updated files (using VsCode Groovy Lint)
- Added .groovylintrc.json so contributors using VsCode + VsCode Groovy Lint ignore the most restrictive violations

Note: locally there is a test failure on SourceFileTest when running gradlew check, as I updated nothing there, I have no idea where it can come from

Related to issue #477

### DIFF
--- a/.groovylintrc.json
+++ b/.groovylintrc.json
@@ -1,0 +1,47 @@
+{
+    "extends": "recommended",
+    "rules": {
+        "CatchThrowable": {
+            "enabled": false
+        },
+        "CompileStatic": {
+            "enabled": false
+        },
+        "DuplicateNumberLiteral": {
+            "enabled": false
+        },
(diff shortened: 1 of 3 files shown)


chrismair:
Please have it print out "CodeNarc " + version, rather than just the raw version. I want the freedom to include more information in that output in the future (like `java -version` or `gradle --version`). 

And please put the `final` modifier after the visibility modifier, so `private final` rather than `final private` (3 occurrences).

Thanks.

nvuillam:
Done :)
I replicated java -version result, without the double quotes around the version, is it ok for you ?

chrismair:
Thanks for the contribution!

PR #485, opened 2020-04-19
Provide results in JSON format

nvuillam:
Provide results in JSON format (same hierarchical structure than XML, except items =[...] for lists)
- New class JsonReportWriter, with associated TestClass JsonReportWriterTest
- Update help, documentation and  add sample json result file


chrismair:
Re: the JSON structure, I propose:
 - Get rid of those “items”. See below.
 - Move “packageSummary” out to be a top-level element and rename it to just “summary”
 - Remove “items” under “packages”. The “packages” element now is just an array of package objects
 - Remove “items” under each “package”. The anonymous package object now includes “files”, whose value is  an array of file objects

So then it looks like:
```
    "codeNarc": { .. }
    "report": { .. }
    "project": { .. }
    "summary": {
        "totalFiles": 6,
        "filesWithViolations": 3,
        "priority1": 2,
        "priority2": 2,
        "priority3": 3
    },
    "packages": [
        {
            "path": "src/main",
            "totalFiles": 3,
            "filesWithViolations": 3,
            "priority1": 2,
            "priority2": 2,
            "priority3": 3,
            "files": [
                {
                    "name": "MyAction.groovy",
                    "violations": [
                        {
                            "ruleName": "RULE1",
                            "priority": 1,
                            "lineNumber": 111,
                            "sourceLine": "if ..",
                            "message": null
                        }, ...
                }
        }
     ]
    "rules": [ .. ]
```

An optional alternative would be to have each package be an object with a single “package” key and object value. But then for consistency, it would seem logical to have a similar “file” and “violation” key/values lower in the hierarchy, rather than the anonymous file and violation objects in your implementation. But that does seem excessive/verbose, and I am inclined to leave it like described above to enable what seems to me to be a more intuitive dot notation access, e.g.:
```
    json.packages[2].path
    json.packages[1].files[3].name
    json.packages[1].files[3].violations[0].ruleName
```

What do you think?

nvuillam:
This is the kind of structure i imagined, i think the one i implemented is too verbose too but i was afraid to change the structure defined in the xml
 i'll make a new proposal :)

nvuillam:
@chrismair , updates are done :)
- report=json:stdout ( I managed only json for stdout option, as I don't think other formats would have a use )
- report=json:somefile.json

New Json output format, with summary, flat package list and no more items
Please notify me if it's ok for u :)

```json
{
    "codeNarc": {
        "url": "https://www.codenarc.org",
        "version": "1.5"
    },
    "report": {
        "timestamp": "Jan 1, 2010 4:51:12 PM"
    },
    "project": {
        "title": "My Cool Project",
        "sourceDirectories": [
            "c:/MyProject/src/main/groovy",
            "c:/MyProject/src/test/groovy"
        ]
    },
    "summary": {
        "totalFiles": 6,
        "filesWithViolations": 3,
        "priority1": 2,
        "priority2": 2,
        "priority3": 3
    },
    "packages": [
        {
            "path": "src/main",
            "totalFiles": 3,
            "filesWithViolations": 3,
            "priority1": 2,
            "priority2": 2,
            "priority3": 3,
            "files": [
                {
                    "name": "MyAction.groovy",
                    "violations": [
                        {
                            "ruleName": "RULE1",
                            "priority": 1,
                            "lineNumber": 111,
                            "sourceLine": "if (count < 23 && index <= 99 && name.contains('\u0000')) {"
                        },
                        {
                            "ruleName": "RULE3",
                            "priority": 3,
                            "lineNumber": 333,
                            "sourceLine": "throw new Exception(\"cdata=<![CDATA[whatever]]>\") // Some very long message 1234567890123456789012345678901234567890",
                            "message": "Other info"
                        },
                        {
                            "ruleName": "RULE3",
                            "priority": 3,
                            "lineNumber": 333,
                            "sourceLine": "throw new Exception(\"cdata=<![CDATA[whatever]]>\") // Some very long message 1234567890123456789012345678901234567890",
                            "message": "Other info"
                        },
                        {
                            "ruleName": "RULE1",
                            "priority": 1,
                            "lineNumber": 111,
                            "sourceLine": "if (count < 23 && index <= 99 && name.contains('\u0000')) {"
                        },
                        {
                            "ruleName": "RULE2",
                            "priority": 2,
                            "lineNumber": 222,
                            "message": "bad stuff: !@#$%^&*()_+<>"
                        }
                    ]
                }
            ]
        },
        {
            "path": "src/main/dao",
            "totalFiles": 2,
            "filesWithViolations": 2,
            "priority1": 0,
            "priority2": 1,
            "priority3": 1,
            "files": [
                {
                    "name": "MyDao.groovy",
                    "violations": [
                        {
                            "ruleName": "RULE3",
                            "priority": 3,
                            "lineNumber": 333,
                            "sourceLine": "throw new Exception(\"cdata=<![CDATA[whatever]]>\") // Some very long message 1234567890123456789012345678901234567890",
                            "message": "Other info"
                        }
                    ]
                },
                {
                    "name": "MyOtherDao.groovy",
                    "violations": [
                        {
                            "ruleName": "RULE2",
                            "priority": 2,
                            "lineNumber": 222,
                            "message": "bad stuff: !@#$%^&*()_+<>"
                        }
                    ]
                }
            ]
        },
        {
            "path": "src/test",
            "totalFiles": 3,
            "filesWithViolations": 0,
            "priority1": 0,
            "priority2": 0,
            "priority3": 0,
            "files": []
        }
    ],
    "rules": [
        {
            "name": "DuplicateImport",
            "description": "Custom: Duplicate imports"
        },
        {
            "name": "UnnecessaryBooleanInstantiation",
            "description": "Use Boolean.valueOf() for variable values or Boolean.TRUE and Boolean.FALSE for constant values instead of calling the Boolean() constructor directly or calling Boolean.valueOf(true) or Boolean.valueOf(false)."
        }
    ]
} 
```

nvuillam:
@chrismair > I'll manage all those points soon, thanks for your feedback :)

nvuillam:
will be reopened later

nvuillam:
> I don’t think it is intuitive that writing to the console would write out the JSON as a single line, or that the report writer would change its format based on whether it is writing to the console. If you need the one-line JSON for your use case, how about adding a boolean property to the report class (perhaps writeAsSingleLine or something like that) and default it to false.

Done with field writeAsSingleLine and value stdout-oneline ( if writeAsSingleLine is not managed on the report class, an IllegalArgumentExcception is thrown )

> I think the JsonReportWriterTest will start failing as soon as I increment the CodeNarc version, right? Try changing the version in “codenarc-version.txt” locally and see if it breaks the test. If it does, see HtmlReportWriterTest @Before method.
> Can you avoid the need to replace/normalize the timestamp in AbstractJsonReportWriterTestCase.normalizeJson() by stubbing out the getFormattedTimestamp() the way that HtmlReportWriterTest does it?

I added variables for version & timestamp  inside hardcoded JSON result to compare


> Is this missing my changes to “codenarc-command-line.md”: remove consoleJson and add my :stdout changes.

Done !

`Are you able to fix the merge conflicts?`

Done !

chrismair:
Nice work. 

I'm sorry I did not think through the need for setting that `writeAsSingleLine` flag on the command-line interface. But I wish you had asked about that * stdout-oneline* option before going all the way down that path. That option just applies to that one report-writer and only for your one use case. I don't think I want to have all that extra logic/docs/checks in the code, and have to support that going forward. 

How about you change your consuming code to use some regex replacement logic to convert the standard fornatted JSON into a one-liner (like the `normalizeXml` method  in `AbstractHtmlReportWriterTestCase`). And then remove all that code from the CodeNarc side.

nvuillam:
I have cases where my consumer code is a call via command line, so i need a way to have one-line json in output just using org.codenarc.Codenarc arguments, and parsing json is way much simpler when it is in one line than when it is prettified, and eventually mixed with debug logs... (Most of CLIs returning json returns it in one line for this reason)

What if i always generate the json in one line , for both file and stdout output ?
If humans wants a more readable result they will probably use the other reports writer like html, or simply indent the json in their favorite text editor ?

chrismair:
Ugh, there is no great answer here, just trying to pick the *lesser evil*. Everything has tradeoffs.

Given that, I suggest adding an `if (reportWriter instanceof JsonReportWriter)` check in `CodeNarc.groovy parseReport()` :
```
    private void parseReport(String argValue) {
        def parts = argValue.split(':', 2)
        def type = parts[0]
        def reportWriter = new ReportWriterFactory().getReportWriter(type)

        if (parts.size() > 1 && parts[1]) {
            if (parts[1] == 'stdout') {
                reportWriter.writeToStandardOut = true

                // Add these three lines ********
                if (reportWriter instanceof JsonReportWriter) {
                    reportWriter.writeAsSingleLine = true
                }
            }
            else {
                reportWriter.outputFile = parts[1]
            }
        }
        reports << reportWriter
    }

}
```

Yes, that is an ugly *special case* that makes me unhappy. But that has the advantages that the ugliness is confined to this one place in one class, and also that it is confined only to the json:stdout case, allowing formatted JSON to be produced from the command-line when writing to files, and also provides formatted JSON to the other 96% of use cases, including the Ant Task.

And then revert all of the other stuff you added to support the more complicated solution.

nvuillam:
I really think that my current dev or the one i proposed (always one line for json) are cleaner than referring JsonReportWriter in CodeNarc class, but... if u prefer this one, i'll makes these changes, the most important for me is to increase perfs of npm-groovy-lint, so this will do :) thanks for you feedback :)

chrismair:
> but... if u prefer this one, i'll makes these changes

Yes, please do. I like those other options even less.

nvuillam:
@chrismair > Rollback done, it should be ok now :)

chrismair:
Almost there.
 - Revert your changes `AbstractReportWriter`. 
 - Add `writeAsSingleLine` only to `JsonReportWriter`. 
 - Update `codenarc-json-report-writer.md` accordingly, to add `writeAsSingleLine`.

nvuillam:
@chrismair done :)

chrismair:
Nice work. Thanks for that contribution!

nvuillam:
You're welcome , open source is made to help each other ^^

PR #499, opened 2020-05-19
allow for annotations before package name

### LINKED ISSUE #498: MissingBlankLineAfterPackage rule fails unexpectedly with package annotations
Hey there friend! 😄 
When playing around with package-level annotations in a `package-info.groovy` file, I ran into a problem with our CodeNarc rules. I found that if I put an annotation above a package like so: 
```
@RandomPackageAnnotation
package com.random.package.of.no.consequence

import com.path.to.RandomPackageAnnotation
```
Then my MissingBlankLineAfterPackage rule would trigger. However, as you can see, there's definitely a blank line after the package haha. After running into this issue, I tried rearranging the lines of code and found that writing the annotation this way:
```
@RandomPackageAnnotation

package com.random.package.of.no.consequence

import com.path.to.RandomPackageAnnotation
```
(with a blank line between the annotation and the package) stopped the MissingBlankLineAfterPackage from complaining. 

I assume the rule makes the assumption that the package will automatically be the top-most line of a file. Either that or the rule is looking for blank lines on either side of the package. In any event, I think this is unexpected behavior...it definitely caught me off guard haha.

jlstrater:
this should resolve #498

### DIFF
--- a/src/main/groovy/org/codenarc/rule/formatting/MissingBlankLineAfterPackageRule.groovy
+++ b/src/main/groovy/org/codenarc/rule/formatting/MissingBlankLineAfterPackageRule.groovy
@@ -34,7 +34,7 @@ class MissingBlankLineAfterPackageRule extends AbstractRule {
     void applyTo(SourceCode sourceCode, List<Violation> violations) {
         PackageNode packageNode = sourceCode.ast?.package
         if (packageNode) {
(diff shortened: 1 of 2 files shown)


chrismair:
Excellent. Thanks for that lightning-quick fix, Jenn.

Issue #502, opened 2020-05-20
SpaceAroundMapEntryColon rule no longer works with characterAfterColonRegex


nre-ableton:
I'm pretty sure that this broke with https://github.com/CodeNarc/CodeNarc/commit/bd78f989438ff4ff538cd82984660c24fb902479 (see https://github.com/CodeNarc/CodeNarc/issues/430), but looking at the diff, I'm not entirely sure _how_ this could happen.

Regardless, since CodeNarc 1.5, the `SpaceAroundMapEntryColon` is throwing all sorts of errors like so:

> `SpaceAroundMapEntryColon: The colon for the literal Map entry for key [ignoreMethodNames] within class None is not followed by a match for regular expression [\\s]`

My (abbreviated) `ruleset.groovy` looks like this:

```groovy
  ruleset('rulesets/formatting.xml') {
    SpaceAroundMapEntryColon(characterAfterColonRegex: /\\s/)
  }
```

I have also tried `SpaceAroundMapEntryColon(characterAfterColonRegex: /\s/)` and even `SpaceAroundMapEntryColon(characterAfterColonRegex: / /)` (with a literal space), but the result is the same in all cases.

chrismair:
Can you provide some example Groovy code that causes these incorrect violations?

I'm pretty sure you want `/\s/` as the regex.

nre-ableton:
Sure, the [Jenkinsfile of groovylint](https://github.com/AbletonDevTools/groovylint/blob/master/Jenkinsfile) is one such example. Here's an excerpt:

```groovy
devToolsProject.run(
  setup: { data ->
    data['venv'] = virtualenv.create('python3.8')
    data.venv.run('pip install -r requirements-dev.txt -r requirements.txt')
  },
  // etc...
)
```

For the above snippet, CodeNarc produces these errors:

> `ERROR ./Jenkinsfile:15: SpaceAroundMapEntryColon: The colon for the literal Map entry for key [setup] within class None is not preceded by a match for regular expression [\S]`
> `ERROR ./Jenkinsfile:15: SpaceAroundMapEntryColon: The colon for the literal Map entry for key [setup] within class None is not followed by a match for regular expression [\s]`

chrismair:
Are you perhaps using Groovy 3.0 or later? I added a test for the above code and it produced no violations when running Groovy 2.4, but produced the above errors running Groovy 3.0.3.

[...]

PR #503, opened 2020-05-20
Fix typo in SpaceAroundMapEntryColon documentation

nre-ableton:


chrismair:
Thanks for the contribution. But it looks like the sample regular expressions are wrong as well.

`/\\S/` should be `/\S/` and `/\\s/` should be `/\\s/`. Those errors are an artifact of our recent migration from APT format to Markdown for the web site. Perhaps that also contributed to the problems in #502.

You can either amend this PR with those fixes or else I'll go in and fix everything at once. Thanks.

nre-ableton:
Sure thing, fixed in a2de5373.

chrismair:
Thanks!

PR #509, opened 2020-06-02
Rule to check for direct usages of GORM methods

musketyr:
It is very difficult to upgrade to newer versions of Grails when you have the persistence logic spread across the whole application. This rule checks for direct usages of GORM instance and static methods so it's easier to find them.

PR #515, opened 2020-06-04
Rule to check for direct usages of GORM methods

musketyr:
It is very difficult to upgrade to newer versions of Grails when you have the persistence logic spread across the whole application. This rule checks for direct usages of GORM instance and static methods so it's easier to find them.

This is reworked #509 excluding the Gradle Wrapper update which I wasn't able to work around.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
+
 # CodeNarc Change Log
 
 TODO: Version 1.6    (?? 2020)
 --------------------------------------
 New Rules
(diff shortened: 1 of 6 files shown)


musketyr:
I'm testing the rule in a sample project and I have weird compilation error happening:

https://github.com/musketyr/grails-codenarc-test

1. publish the code from this branch as `1.6` version locally with `./gradlew install`
2. run the analysis in the demo repo `./gradlew codenarcMain`

chrismair:
Thanks for the contribution.

I could not reproduce the `./gradlew install` on your project. If there is a real error, please open a separate issue.

musketyr:
Thank you too!

PR #518, opened 2020-06-12
Fix #517: MissingBlankLineAfterImportsRule: add support for all import types

### LINKED ISSUE #517: MissingBlankLineAfterImportsRule does not support the 4 different kinds of imports
For example a star import with a correct blank line after the import is reported as violation.
```
package org.codenarc
            
import org.codenarc.rule.Rule
import java.lang.*

class MyClass { }
```

darxriggs:
The implementation formerly only considered normal imports.

Clean up tests & make them more consistent too.


chrismair:
Nice. Thanks for the contribution.

PR #520, opened 2020-06-12
Update publishing mechanism

darxriggs:
This one is up for discussion. I suggest that you try it locally before merging. Because I don't know the details of the current release process, I introduced the "release" project property (according to some Gradle documentation) to decide whether to publish to the "release" or the "snapshot" repository.

The following text is from the commit comment:

The "maven" plugin is deprecated. The "maven-publish" plugin should be used.

Gradle module metadata is now published too.

The following tasks are now relevant for publishing:
- publishToMavenLocal - Publishes all Maven publications produced by this project to the local Maven cache.
- publish - Publishes all publications produced by this project.

If the project property "release" is present, publishing is using the "release" instead of the "snapshot" repository (`./gradlew -Prelease publish`).

References
- https://docs.gradle.org/6.5/userguide/upgrading_version_5.html#changes_6.0
- https://docs.gradle.org/6.5/userguide/publishing_setup.html
- https://docs.gradle.org/6.5/userguide/publishing_maven.html
- https://docs.gradle.org/6.5/userguide/java_plugin.html
- https://docs.gradle.org/6.5/userguide/signing_plugin.html
- https://blog.gradle.org/gradle-metadata-1.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id 'groovy'
-    id 'maven'
+    id 'maven-publish'
     id 'signing'
     id 'idea'
 }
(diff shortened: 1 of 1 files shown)


chrismair:
Nice. Looks promising. It may be a couple weeks before we can get to this. Thanks.

chrismair:
Works great! Very helpful contribution. Thanks.

It turns out that for publishing to Sonatype, the version must end in “SNAPSHOT” to be published to the snapshot repo. So I will probably switch the `maven url` logic to go off that instead.

Issue #510, opened 2020-06-02
Improving indentation support for chained methods with closure parameters


dmurat:
Currently, support for chained methods does not account for methods with closure bodies. For example, if the following test is added in `IndentationRuleTest`, the test will fail on `println someName` statement:

    @Test
    void test_MethodChaining_MultilineClosureParameter() {
        final SOURCE = '''
            |buildFileList().collect { item -> item.name }
            |               .each { name -> println name }
            |               .each { someName -> 
            |                   println someName
            |               }
            |               .each { name -> println name }
            |
        '''.stripMargin()
        assertNoViolations(SOURCE)
    }

Is it possible to add this support?

Tnx.

chrismair:
I think the level of indentation for the .each lines in your example is off by one (or five or nine) columns.

In any case, the changes for the upcoming CodeNarc 1.6 have made this rule more lenient/robust. So this now works (jn the master branch):
```
    void test_MethodChaining_MultilineClosureParameter() {
        final SOURCE = '''
            |buildFileList().collect { item -> item.name }
            |        .each { name -> println name }
            |        .each { someName ->
            |            println someName
            |        }
            |        .each { name -> println name }
        '''.stripMargin()
        assertNoViolations(SOURCE)
    }
```
as does the same code with only a single level of indentation for the .each.

dmurat:
Tnx for looking into this. I think it would be great if CodeNarc can support some common styles of chained methods indentation. I think that style 1 and 2 in the following example are most frequently used in the wild. At least in code that cares enough about indentation 😃 

```
// style 1
buildFileList()
        .collect { item -> item.name }
        .each { name -> println name }
        .each { someName -> 
            println someName
        }
        .each { name -> println name }

// style 2
buildFileList().collect { item -> item.name }
               .each { name -> println name }
               .each { someName -> 
                   println someName
               }
               .each { name -> println name }

// style 3
buildFileList().collect { item -> item.name }
        .each { name -> println name }
        .each { someName ->
            println someName
        }
        .each { name -> println name }
```
I didn't see much of style 3 since it visually hides the first method in the chain.

CodeNarc currently supports style 1 and 3, but style 2 fails with an indentation rule violation. I realize that Indentation rule implementation is pretty complex, and I'm having rather hard time following it. I guess this is one of the reasons for some inconsistencies that I stumbled upon. This might be seen in the next test example:

```
    @Test
[...]

PR #536, opened 2020-07-30
Method chaining indentation improvements

dmurat:
This PR implements https://github.com/CodeNarc/CodeNarc/issues/510

At the moment CodeNarc's IndentationRule supports the indentation checks for following two styles of chained methods:

```
// style 1
buildFileList().collect { item ->
            item.name
        }
        .each1 { name -> println name }
        .each2 { someName ->
            println someName
        }
        .each3 { name ->
            println name
        }

// style 2
buildFileList()
        .collect { item ->
            item.name
        }
        .each1 { name -> println name }
        .each2 { someName ->
            println someName
        }
        .each3 { name ->
            println name
        }

```

This PR adds support for additional style of chained methods indentation:
```
// style 3
buildFileList().collect { item ->
                   item.name
               }
               .each1 { name -> println name }
               .each2 { someName ->
                   println someName
               }
               .each3 { name ->
                   println name
               }
```

In addition, PR also improves the reporting of expected columns for chained method calls.

Tnx


chrismair:
Nice work! A few requests:

 - Rename the test class to `IndentationRule_MethodChainingTest` to follow naming convention for other "extra" test classes. And change @author to you.
 - Provide brief textual description of style 1, 2, 3 (and examples if necessary) in the comments at the top of the test class. I am confused because style 2 and 3 seem reversed from how they are described in #510.
- Please locally test with Groovy 3 and make sure the two `IndentationRule` test classes run clean. That rule is especially problematic in Groovy 3. For instance, temporarily modify `build.gradle`:
```
        def groovyVersion = '3.0.4'
        //def groovyVersion = '2.5.12'
```

dmurat:
Tnx. 

Yes, you are right. Styles references in PR and in #510 are different. In PR I started with what already exists in CodeNarc (style 1 and 2), and the added new supported style 3.

I've just run local tests with Groovy 3.0.5. `IndentationRuleTest` and `IndentationRule_MethodChainingTest` passes cleanly. However, `UnnecessarySemicolonRuleTest.test_Field()` fails. This is the only test (of 5267) that fails with Groovy 3.0.5.
As it seems unrelated, I didn't check the details. Is this ok?

dmurat:
To make sure that my changes did not mess up something, as an additional check, I just run all tests against the `master` branch with Groovy 3.0.5. `UnnecessarySemicolonRuleTest.test_Field()` fails in that case too with the same failure output. So it really looks like it is not related to this PR.

chrismair:
Thanks for the contribution! And thanks for doing the extra testing with Groovy 3. Yeah, supporting both Groovy 2.x and 3.x will be a bit of a pain for a while.

Issue #548, opened 2020-08-24
New OptionalField rule: Do not use an Optional as a field type


chrismair:
Do not declare a field of type Optional.

chrismair:
Available in CodeNarc 2.0.0.

PR #552, opened 2020-09-07
JSON Rulesets management + XML/JSON as string arguments

nvuillam:
- #548: Allow to use RuleSets in JSON format ([Nicolas Vuillamy](https://github.com/nvuillam))
 - #548: Allow to send XML and JSON RuleSets as string arguments ([Nicolas Vuillamy](https://github.com/nvuillam))

+ some source formatting according to some CodeNarc rules 
+ some markdown formatting using markdownlint
+ changelog

Notes:
- I'll use that from npm-groovy-lint, to avoid to have to generate temporary ruleSetFiles but directly send the string as JSON, so performances will be again improved 🚀 
- I have no use of the XML string sent as argument, but it would have been sad to manage only my requirement, so I also implemented it 👼


nvuillam:
For some reason, the CI job is not displayed here, but he  has run correctly 🏃 

-> https://travis-ci.org/github/CodeNarc/CodeNarc/builds/725076506

chrismair:
Is it too late to split this into two separate PRs? This is massive, and implements two separate, standalone capabilities.

nvuillam:
I did them together because I would have needed the first one to be merged before being able to develop the second one , and I developed both in a single evening... :)

If it's really important for you, i can :
- make a new PR with Json format management
- wait for you to merge
- make a new PR with rulesets as string arguments (Json/XML)

Plz keep me posted :)

nvuillam:
Replaced by #554  :) 
... then by another one with XML/JSON as string argument when 554 will be merged :)

chrismair:
For the new PR to support rulesets as strings (JSON/XML), I think rather than overloading the `rulesetfiles` parameter, it would be better to just introduce a new `ruleset` parameter, and that would take precedence if present.

PR #555, opened 2020-09-10
Add GitHub Action Super-Linter

nvuillam:
https://github.com/marketplace/actions/super-linter

DRAFT PR

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,53 @@
+---
+###########################
+###########################
+## Linter GitHub Actions ##
+###########################
+###########################
+name: Lint Code Base
+
+#
+# Documentation:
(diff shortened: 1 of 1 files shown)


nvuillam:
mmmm not working , probably because i'm from a fork

@chrismair maybe you could give me some rights so I can create a branch directly in the main repository ? ( with no write access to master so you won't be afraid I break something 👼  )

chrismair:
I'm not inclined to hook this in, at least for now. It looks quite cool and helpful in general, but we already have properly customized CodeNarc validation running as part of the test suite. And adding more external tools/libraries/frameworks always results in work and pain later when they break or must be upgraded.

nvuillam:
It was just an attempt, for other languages/formats than groovy (and also because i contribute to Super-Linter 🤓) , but the refactoring would take too much time for low benefit, i close it :)

PR #568, opened 2020-11-06
Create a Docker image

ericcitaire:
* add fatJar task to Gradle build
* add Dockerfile
* customize Docker Hub build

An image is already available on Docker Hub as [`ericcitaire/codenarc`](https://hub.docker.com/repository/docker/ericcitaire/codenarc).

Example usage :
```
docker run --rm -v $(pwd):/code ericcitaire/codenarc:2.0.0 codenarc -report=text:stdout
```

This can be really useful in CI builds for example.

Cheers,
Eric.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,44 @@
-FROM openjdk:8-jdk-slim AS build
+ARG CODENARC_VERSION=2.0.0
+ARG JAVA_VERSION=8
+ARG GROOVY_VERSION=2.5.12
+ARG SLF4J_VERSION=1.7.25
+ARG GMETRICS_VERSION=1.1
 
-WORKDIR /code
+FROM buildpack-deps:stable-curl AS downloads
 
-COPY . /code
+ARG GROOVY_VERSION
+ARG CODENARC_VERSION
+ARG SLF4J_VERSION
+ARG GMETRICS_VERSION
 
-RUN ./gradlew --quiet --no-daemon shadowJar
+RUN mkdir -p /downloads && cd /downloads \
+ && curl -fsSLO https://search.maven.org/remotecontent?filepath=org/codenarc/CodeNarc/$CODENARC_VERSION/CodeNarc-$CODENARC_VERSION.jar \
+ && curl -fsSLO https://search.maven.org/remotecontent?filepath=org/codehaus/groovy/groovy/$GROOVY_VERSION/groovy-$GROOVY_VERSION.jar \
+ && curl -fsSLO https://search.maven.org/remotecontent?filepath=org/codehaus/groovy/groovy-xml/$GROOVY_VERSION/groovy-xml-$GROOVY_VERSION.jar \
+ && curl -fsSLO https://search.maven.org/remotecontent?filepath=org/codehaus/groovy/groovy-json/$GROOVY_VERSION/groovy-json-$GROOVY_VERSION.jar \
+ && curl -fsSLO https://search.maven.org/remotecontent?filepath=org/codehaus/groovy/groovy-ant/$GROOVY_VERSION/groovy-ant-$GROOVY_VERSION.jar \
+ && curl -fsSLO https://search.maven.org/remotecontent?filepath=org/codehaus/groovy/groovy-templates/$GROOVY_VERSION/groovy-templates-$GROOVY_VERSION.jar \
+ && curl -fsSLO https://search.maven.org/remotecontent?filepath=org/slf4j/slf4j-api/$SLF4J_VERSION/slf4j-api-$SLF4J_VERSION.jar \
+ && curl -fsSLO https://search.maven.org/remotecontent?filepath=org/slf4j/slf4j-simple/$SLF4J_VERSION/slf4j-simple-$SLF4J_VERSION.jar \
+ && curl -fsSLO https://search.maven.org/remotecontent?filepath=org/gmetrics/GMetrics/$GMETRICS_VERSION/GMetrics-$GMETRICS_VERSION.jar
 
-RUN echo '#!/bin/bash' > codenarc \
- && echo '/usr/local/openjdk-8/bin/java -jar /opt/codenarc.jar "$@"' >> codenarc \
- && chmod 755 codenarc
+FROM openjdk:$JAVA_VERSION
 
-FROM openjdk:8-jre-slim
+ARG GROOVY_VERSION
+ARG CODENARC_VERSION
+ARG SLF4J_VERSION
+ARG GMETRICS_VERSION
 
-COPY --from=build /code/build/libs/CodeNarc-*-all.jar /opt/codenarc.jar
-COPY --from=build /code/codenarc /usr/local/bin/codenarc
+ENV GROOVY_VERSION=$GROOVY_VERSION
+ENV CODENARC_VERSION=$CODENARC_VERSION
+ENV SLF4J_VERSION=$SLF4J_VERSION
+ENV GMETRICS_VERSION=$GMETRICS_VERSION
+
+COPY --from=downloads /downloads/*.jar /opt/
+
+RUN echo '#!/bin/bash' > /usr/local/bin/codenarc \
+ && echo "$JAVA_HOME/bin/java -cp '/opt/CodeNarc-$CODENARC_VERSION.jar:/opt/groovy-$GROOVY_VERSION.jar:/opt/groovy-xml-$GROOVY_VERSION.jar:/opt/groovy-json-$GROOVY_VERSION.jar:/opt/groovy-ant-$GROOVY_VERSION.jar:/opt/groovy-templates-$GROOVY_VERSION.jar:/opt/slf4j-api-$SLF4J_VERSION.jar:/opt/slf4j-simple-$SLF4J_VERSION.jar:/opt/GMetrics-$GMETRICS_VERSION.jar'"' org.codenarc.CodeNarc "$@"' >> /usr/local/bin/codenarc \
+ && chmod 755 /usr/local/bin/codenarc
 
 WORKDIR /code
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM openjdk:8-jdk-slim AS build
+
+WORKDIR /code
+
+COPY . /code
+
+RUN ./gradlew --quiet --no-daemon shadowJar
+
+RUN echo '#!/bin/bash' > codenarc \
+ && echo '/usr/local/openjdk-8/bin/java -jar /opt/codenarc.jar "$@"' >> codenarc \
+ && chmod 755 codenarc
+
+FROM openjdk:8-jre-slim
+
+COPY --from=build /code/build/libs/CodeNarc-*-all.jar /opt/codenarc.jar
+COPY --from=build /code/codenarc /usr/local/bin/codenarc
+
+WORKDIR /code
+
+CMD /usr/local/bin/codenarc
--- a/hooks/post_push
+++ b/hooks/post_push
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+VERSION=$(docker run --rm "$IMAGE_NAME" codenarc -version | sed 's/^CodeNarc version //')
+
+docker tag "$IMAGE_NAME" "$DOCKER_REPO:$VERSION"
+docker push "$DOCKER_REPO:$VERSION"


nvuillam:
FYO, CodeNarc is already available behind [npm-groovy-lint](https://github.com/nvuillam/npm-groovy-lint) in a [docker image](https://github.com/nvuillam/npm-groovy-lint#call-via-docker), or event in [Mega-Linter](https://github.com/nvuillam/mega-linter#readme) , with all other linters preconfigured  :)

ericcitaire:
> FYI, CodeNarc is already available behind [npm-groovy-lint](https://github.com/nvuillam/npm-groovy-lint) in a [docker image](https://github.com/nvuillam/npm-groovy-lint#call-via-docker), or even in [Mega-Linter](https://github.com/nvuillam/mega-linter#readme) , with all other linters preconfigured :)

@nvuillam Thans for the links. I did notice the existence of these 2 images, but I have some concerns about it.

I don't really see the point of using NPM packaging for a Groovy linter written in pure Groovy / Java. It just doesn't feel right.

I understand that it might make sense in the specific case of the VSCode extension, but in the general case, Node / NPM doesn't add any value to it.

IMHO, CodeNarc should be packaged as a standalone fat JAR, not a NPM package.

nvuillam:
I understand that the mix node/npm + Java/Groovy can seem surprising 🤣 

But there is a story... I initially just packaged CodeNarc to be distributed on NPM using [jdeploy](https://github.com/shannah/jdeploy) because the "Command line" use of CodeNarc required to write long command, download jar dependencies, define CLASSPATH... and I wanted a one-liner to have a CodeNarc ready to use out of the box as CLI

- Then corona happened, I had time .... and I refactored arguments , added default config, so they are more user friendly
- Then created a CodeNarc Server to avoid to launch JVM at each CodeNarc call
- Then added the capability to auto-format and auto-fix
- Then contributed to CodeNarc to add some useful capabilities ...
- Then embedded it in a VsCode extension :)

So ... not the fruit of a long reflexion before, but more an incremental dev to fit user requests :)

ericcitaire:
@nvuillam Just to be clear : you did a great job on `npm-groovy-lint` and I thank you for this. You probably had very good reasons to do it this way, I understand it. But I think we need to step back and think about it.

I want to use CodeNarc because it seems to be the right tool for the job. Unfortunately, it lacks some features. Just like you, I want a one-liner command to check my Groovy code.

Yes, `npm-groovy-lint` fills the gap. But can I trust it? Nothing personal here, but as a developer, I need to trust my tools.

Here are some of my concerns :

* it is not part of the *official* project
* it is built from non-official dependencies JARs (see [`lib/java`](https://github.com/nvuillam/npm-groovy-lint/tree/master/lib/java))
* dependencies versions are different from the official project (example : [Groovy 2.5](https://github.com/CodeNarc/CodeNarc/blob/master/build.gradle#L20) vs [Groovy 3.0](https://github.com/nvuillam/npm-groovy-lint/tree/master/lib/java/groovy/lib))
* it uses NPM, which I don't need (and TBH, which I don't really trust...)

IMO, your work (default config, CodeNarc Server, auto-format, auto-fix, etc.) is of great value and should be part of the official project. It would be much appreciated.

But about the one-liner command, I think a fat JAR is the way to go.

nvuillam:
I understand your points, the day CodeNarc can natively support what brings npm-groovy-lint, I'll be ok to deprecate it, it will be less maintenance for me 👍 
But before that, I assure you that you can trust npm-groovy-lint :) ( except that i've not upgraded it to CodeNarc v2 yet... will do someday )

erdi:
Hey @ericcitaire - I'd be interested in your motivation for submitting this PR. Is it to provide an executable JAR thanks to which you will be able to run CodeNarc from the command line easily? If that's the case then why not make the result of the `fatJar` task that you added to the build to be part of publishing so that it's uploaded to maven when a new version of CodeNarc is released? I don't know if @chrismair would be willing to accept what I'm suggesting but I have to admit that I don't see a point in involving Docker in the process - I don't understand the drive to wrap every single JVM based library in a Docker image that I've been observing of late. Unless there is a reason for distributing a Docker image with JDK8 and CodeNarc that I'm not seeing?

ericcitaire:
@erdi Yes, providing a standalone executable jar to easily run CodeNarc from the command-line is one of my motivations. I agree that it would be nice to publish it to Maven central, but I'm more confortable with Maven than Gradle, and I just know how to do it... I will find out how to do it.
Code linting is usually part of Continuous Integration. Providing a Docker image helps with CI tools like Jenkins, Gitlab CI, GitHub Actions, etc. where you can run your verification process in one or more containers.

ericcitaire:
@erdi I made some edits to `build.gradle` so that the fat JAR is now part of publishing as you suggested.
Again, I'm not an expert with Gradle. There might be a better way to do it. Let me know.

chrismair:
Okay, I'll take a look. Publishing that fatJar seems like a reasonable idea. 

Is that Docker config generic/universal enough that it would be "the" way that most folks would configure it to run CodeNarc on Docker? I would not want to start getting requests to support different "flavors" and have to maintain customizations or variations.  Is there significant advantage to including that within the CodeNarc project vs just having it be a separate config/setup/artifact?

ericcitaire:
@chrismair 

> Is that Docker config generic/universal enough that it would be "the" way that most folks would configure it to run CodeNarc on Docker? I would not want to start getting requests to support different "flavors" and have to maintain customizations or variations.

I think it's a very good starting point. It might be nice to provide different versions of Java, each one having its own image tag. We can do this by adding a build arg to the image and by customizing the Docker Hub automated build hooks.

> Is there significant advantage to including that within the CodeNarc project vs just having it be a separate config/setup/artifact?

Yes. Including the Dockerfile in the project allows to build an up-to-date image every time there is a change to the code. It can be automatically built by Docker Hub on each push on master branch.

build.gradle
```diff
@@ -5,6 +5,17 @@ plugins {
     id 'idea'
 }
 
+task fatJar(type: Jar) {
```
erdi, line 8: I don't think we should be explicitly adding all this configuration if there are plugins that will do all that for us, like the [shadow plugin](https://plugins.gradle.org/plugin/com.github.johnrengelman.shadow) which is believe is the industry standard for building fat jars. If you add:
```
id 'com.github.johnrengelman.shadow' version '6.1.0'
```
to the `plugins {}` block at the top of this file and then configure the main class:
```
shadowJar {
    manifest {
        attributes 'Main-Class': 'org.codenarc.CodeNarc'
    }
}
```
then that will achieve all that we need, including adding the fat jar to the publication and all of the changes you made could be reverted. I find that approach simpler and more idiomatic.

erdi:
I have to say that adding publishing of a fat jar feels valuable to me, especially given that it's very simple if we use Gradle Shadow plugin and should require fairly low maintenance if any.

On the other hand I'm struggling to see the benefit of wrapping CodeNarc in a docker image. I personally would never use CodeNarc via a dedicated docker image. Given that CodeNarc is part of the jvm ecosystem it feels to me like the more common way to use CodeNarc is via build tools from that ecosystem like Gradle or Maven. I'm really struggling to see why when building a project using Groovy on CI you would not have a jvm based build tool in place and why you would not use that to perform CodeNarc analysis and instead reach out to a docker image.

> Providing a Docker image helps with CI tools like Jenkins, Gitlab CI, GitHub Actions, etc. where you can run your verification process in one or more containers.

But would that not lead to duplication where locally you would run the verification using a build tool and you would do it differently on CI? Wouldn't you need to keep version of CodeNarc in sync in two places - one for CI and one for building locally, @ericcitaire? Or would you also run CodeNarc in docker locally?

FWIW, @chrismair, I would personally not make publishing of a docker image part of the core project. Don't get me wrong, I'm using docker heavily in CI (e.g. https://github.com/geb/geb/blob/master/.circleci/config.yml#L5) and I believe that running CI builds in docker containers gives you amazing portability between various CI servers and providers. But I use it differently - the CI image only has the JDK and any non-jvm based dependencies and all of the jvm based dependencies are handled and resolved by the build system, the same way as they are when running locally.

ericcitaire:
@erdi @chrismair Obviously, introducing a Docker image build seems to be an issue for you. While I don't fully understand why, I would like to submit a separate PR dedicated to the fat jar, since you both agree on its added value. Would it help to move forward on this?

erdi:
> While I don't fully understand why

I believe I have already explained my problems with this but I'm going to summarise anyway:
- I don't see the benefit of wrapping a java library/tool into a docker image when there are more conventional ways of consuming it (build tools) in CI
- every addition to a project has to be maintained going forward - it's hard to take this on if you consider the addition to have a very limited audience

In my opinion (and I'm just a contributor and not a maintainer so I might be speaking out of line) it's best if you stick to publishing these images at [`ericcitaire/codenarc`](https://hub.docker.com/r/ericcitaire/codenarc) if they are useful to you.

build.gradle
ericcitaire: Done.

ericcitaire:
@erdi 

> * I don't see the benefit of wrapping a java library/tool into a docker image when there are more conventional ways of consuming it (build tools) in CI

Not every project needs a build tool like Maven or Gradle, especially when there is nothing to build.

In my specific use case, I want to add some validation on a bunch of Groovy scripts meant to be used by Jenkins (Job DSL scripts). These scripts are not meant to be compiled, packaged or published, they are just plain scripts, hosted on a Git repository and consumed as is.

> * every addition to a project has to be maintained going forward - it's hard to take this on if you consider the addition to have a very limited audience

I don't think that the audience is *very limited*, I think this is quite the opposite.

There are [plenty of CodeNarc Docker images out there](https://hub.docker.com/search?q=codenarc&type=image). Have a look at [`edupo/codenarc` (100K+ pulls)](https://hub.docker.com/r/edupo/codenarc), [`abletonag/groovylint` (50K+ pulls)](https://hub.docker.com/r/abletonag/groovylint), [`nvuillam/npm-groovy-lint` (1.6K pulls)](https://hub.docker.com/r/nvuillam/npm-groovy-lint).

Some random guys (including me) are building their own image because there is no official one. These unofficial builds are often out of date. And as I mentioned [earlier](https://github.com/CodeNarc/CodeNarc/pull/568#issuecomment-722989845), it is difficult to have trust in these non-official images.

nvuillam:
Agreed with @ericcitaire : The 100k pulls image is using... CodeNarc 1.1 🤣 

And official image would be better :) (or at least a link to an image containing an up-to-date and maintained CodeNarc ... 👼  )

erdi:
> I want to add some validation on a bunch of Groovy scripts meant to be used by Jenkins (Job DSL scripts)

I wish you mentioned that earlier, @ericcitaire. This is indeed a reasonable use case where there is no build tool involved and using a Docker image is more convenient.

> Have a look at edupo/codenarc (100K+ pulls)

Ok, that proves that there are people needing it. Thanks for sharing.

I'm still somewhat skeptical about the fact that such a docker image will tie the versions of Java and Groovy (especially Groovy cause you usually want to use the same version of Groovy for CodeNarc analysis as you use for compiling your code). And I would argue that publishing an image on each code push is and overkill (will people really be using snapshot versions of CodeNarc in this way?) - I think that publishing only for final versions will be enough. For that we could [publish from tags only](https://docs.docker.com/docker-hub/builds/#set-up-build-rules). Finally I think that we should not be building the fat jar as part of building the image but we should just download it from maven central.

These are my final 2 cents. I will shut up now. :)

chrismair:
Interesting discussion. Thank you all for the input.

I agree with @erdi that I would not want to publish a new image with each push to master, but would want it done only for versions/tags. 

And I am still wondering about the best way to handle the expanding set of variations for Java + Groovy versions.

ericcitaire:
@chrismair 

> I agree with @erdi that I would not want to publish a new image with each push to master, but would want it done only for versions/tags.

Yes, of course. It can easily be done by configuring the automated build on Docker Hub to be triggered on tag push.
 
> And I am still wondering about the best way to handle the expanding set of variations for Java + Groovy versions.

It can definitely be done by adding build arguments for Java and Groovy versions.

I can work on it and update the PR accordingly.

@erdi 

> I think that we should not be building the fat jar as part of building the image but we should just download it from maven central.

I'm afraid this will be very difficult if not impossible. The Docker build will be triggered before the artifacts will be available on Maven Central.

Also, the fat jar will contain the version of Groovy specified in `build.gradle`. That will prevent us to build each variations of the image, unless we build it with the normal jar.

ericcitaire:
I get some test failures with Groovy 3.0.6. Can you help me on this ? or should I skip the tests entirely while building the Docker image ?

---

Code using `@Grab` fails to be parsed due to missing Ivy classes :

```
org.codenarc.rule.ClassResolutionTest > testGrabError FAILED
    Assertion failed: 

    assert sourceCode.valid
           |          |
           |          false
           SourceString[
                       @Grab(group='org.springframework', module='spring', version='2.5.6')
                       import org.springframework.jdbc.core.JdbcTemplate
                   ]
```

After adding the Ivy dependency explicitly in `build.gradle` the test passes.

---

For some reason, the AST occasionally returns additional parenthesis on expressions :

```
org.codenarc.rule.unnecessary.UnnecessarySetterRuleTest > testSingleViolation FAILED
    java.lang.AssertionError: Violation on line 2
    Expected message text: [setProperty(1 + 2) can probably be rewritten as property = 1 + 2]
    Found message text:    [setProperty((1 + 2)) can probably be rewritten as property = (1 + 2)]
```

```
org.codenarc.rule.unnecessary.UnnecessarySetterRuleTest > testMultipleViolations FAILED
    java.lang.AssertionError: Violation on line 2
    Expected message text: [setProperty(1) can probably be rewritten as property = 1]
    Found message text:    [setProperty((1)) can probably be rewritten as property = (1)]
```

chrismair:
Ugh, Groovy 3.x has been a continuing challenge. It seems like each point release brings a few broken tests and/or minor AST changes. I have opened a separate issue for those test failures: #571 

That being said, I still wonder whether this Docker issue would be better handled as a simpler Docker build that just pulls down the versioned CodeNarc fat jar, and just do it as a quick manual step after each release is in Maven Central repo.

ericcitaire:
> That being said, I still wonder whether this Docker issue would be better handled as a simpler Docker build that just pulls down the versioned CodeNarc fat jar, and just do it as a quick manual step after each release is in Maven Central repo.

Ok, let's start with this. I'll modify the `Dockerfile` to do exactly that and add build args for Java and Groovy versions.

chrismair:
That sounds good to me. Thanks.

ericcitaire:
@chrismair I've rewritten the `Dockerfile` to allow different versions of Java and Groovy.

Here is an example of build command :

```
JAVA_VERSION=11 GROOVY_VERSION=3.0.6 docker build \
  --build-arg JAVA_VERSION --build-arg GROOVY_VERSION \
  --tag codenarc/codenarc:2-openjdk-8-groovy-3 .
```

I'm not really sure about the Classpath. Transitive dependencies are missing, it might be an issue.
It would have been a lot easier to do this with Gradle. That's why I made this in a separate commit, so we can change our mind. :)

ericcitaire:
An alternative would be to add a Gradle property to override the Groovy version, like this :

```
def groovyVersion = project.hasProperty('groovyVersion') ? project.groovyVersion : '2.5.12'
```

and use it like this :

```
./gradlew shadowJar -PgroovyVersion=3.0.6
```

This implies to go back to the previous way of building the Docker image.

The main disadvantage is that the jar contained in the Docker image is not the *official* one published on Maven Central. On the other hand, this would be much easier to maintain since every change to the Gradle build would be reflected in the Docker image without any effort.

chrismair:
Thanks for all that work. Yeah, I have to admit that is more involved than I was hoping. But I was expecting we would just pull down the fatjar, though I know that fatjar has not yet been published. Once that is published, won't this be much simpler? Should I go ahead and publish a 2.0.1 jar just to get a fatjar out there?

ericcitaire:
@chrismair The published fatjar won't help us, because it already contains Groovy 2.5.12.

I suggest to build the fatjar in-place with a Gradle property.

ericcitaire:
@chrismair The published fatjar won't help us, because it already contains Groovy 2.5.12.

I suggest to build the fatjar in-place with a Gradle property.

chrismair:
I've spent some time playing around with Docker + Gradle, having Docker generate and invoke a Gradle script to pull down a specified CodeNarc and CodeNarc versions, along with the required dependencies. I got that part working fine, but so far could not figure out passing command-line parameters seamlessly from `docker run` --> Gradle task . I would expect that overall approach is probably feasible, but it was not "simple", and running it through Gradle brings its own unpleasantness. But there may be alternatives to running it through Gradle, e.g. creating a ShadowJar at Docker build time and just running that - I gratefully acknowledge that is similar to your proposal(s) as well, and borrows several ideas from them.

I'd like to spend some more time playing around with that, and if I am able to get it working, then evaluate the various options. There seems to be no *perfect* option here.

ericcitaire:
@chrismair You can also have a look at [Google Jib](https://github.com/GoogleContainerTools/jib) and its [Gradle plugin](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin). It's really good at building Java programs images. It doesn't require Docker and optimises image layers by separating the dependencies and the application itself.

Be careful to **not use** the default base image. These distro-less images do not contain anything but Java itself (no shell, no nothing). While this can be appropriate for a production app by reducing its attack surface and its size, this can be really annoying for a development tool like CodeNarc.

I'll try to play with it as well and submit the code if I'm successful with it.

chrismair:
I got the Docker + Gradle + Shadowjar working. See #573.

I think I do prefer an approach that uses the formal versioned CodeNarc jar, since we know that is feasible.

I am not thrilled about having to pull down Gradle in #573 just to resolve/download the CodeNarc transitive dependencies. I am still debating the pros and cons of that versus the more explicit approach in this PR. 

Feedback and opinions are welcome.

chrismair:
I have merged #573. That approach was heavily influenced by the code and discussion in this issue.

I believe that addresses the requirements in this issue, so if there is no further discussion, I'll close this one. Feel free to reopen if necessary. Thanks.

PR #569, opened 2020-11-09
Adding the Inline Console Report Writer to facilitate the read using format strings like errorformat

luiszimmermann:
With this simple format we can read the output easily with grok and other types of format strings.

Tests heavily inspired by TextReportWriter.

### DIFF
--- a/src/main/groovy/org/codenarc/report/InlineConsoleReportWriter.groovy
+++ b/src/main/groovy/org/codenarc/report/InlineConsoleReportWriter.groovy
@@ -18,7 +18,6 @@ package org.codenarc.report
 import org.codenarc.AnalysisContext
 import org.codenarc.results.FileResults
 import org.codenarc.results.Results
-import org.codenarc.rule.Violation
 
 /**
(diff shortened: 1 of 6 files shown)


chrismair:
Nice contribution. I don't see the need for a separate `AbstractInlineConsoleReportWriterTestCase` class, and it is very similar to `AbstractTextReportWriterTestCase`. Can you just inherit from `AbstractTextReportWriterTestCase` instead?

chrismair:
Also, I think the report writer class name might be somewhat confusing or misleading -- both the words "console" and "inline". The `InlineXmlReportWriter` produces a report with the descriptions *inline*. And this is another "text" report, so it seems like "console" just adds another word for the same thing. How about `CompactTextReportWriter` and then `ReportWriterFactory` can just use the short name "compact"?

chrismair:
@luiszimmermann: are you planning on making the changes I suggested?

chrismair:
I merged and also renamed "inlineConsole" report name to "compact", and renamed `InlineConsoleReportWriter` to `CompactTextReportWriter`.

PR #573, opened 2020-11-27
Dockerfile to build CodeNarc Docker image using Gradle and Shadowjar

chrismair:
Dockerfile to build CodeNarc Docker image using Gradle and Shadowjar, allowing specification of Groovy and CodeNarc versions.

An alternative to the options presented in #568.

To build the image:
```
docker build -t codenarc .
# or
docker build -t codenarc --build-arg CODENARC_VERSION=1.6.1 --build-arg GROOVY_VERSION=3.0.6 .
```

To run (assumes there is a "codenarc.ruleset" file in the current directory):
```
docker run --rm -v $(pwd):/work --user $(id -u):$(id -g) codenarc -rulesetfiles=file:codenarc.ruleset
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,10 +1,10 @@
-FROM openjdk:8-jdk
+FROM openjdk:8-jre-slim AS gradle
 
 ARG CODENARC_VERSION=2.0.0
 ARG GROOVY_VERSION=2.5.12
 ARG GRADLE_VERSION=6.5.1
 
-RUN apt-get update && apt-get install -y unzip
+RUN apt-get update && apt-get install -y unzip curl
 WORKDIR /gradle
 RUN curl -L https://services.gradle.org/distributions/gradle-$GRADLE_VERSION-bin.zip -o gradle-$GRADLE_VERSION-bin.zip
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN gradle --quiet --no-daemon shadowJar
+
+FROM openjdk:8-jre-slim
+
+WORKDIR /lib
+COPY --from=gradle /gradle/build/libs/codenarc-all.jar /lib/codenarc-all.jar
+RUN chmod 755 /lib/codenarc-all.jar
+
+WORKDIR /work
+COPY . /work
```
erdi, line 26: I don't believe this `COPY` command is necessary. As far as I understand it copies contents of the directory from which the image is build (`/docker`) into `/work` at image build time. I think there is anything of value in that directory for executing codenarc as `/work` is the dir where the execution happens if I understand correctly.

erdi:
Looks like a sensible approach to me.

docker/Dockerfile
chrismair: Agreed. I have removes that `COPY`. Thanks.

PR #576, opened 2020-12-18
Ignore pointcuts for empty methods

tlefevre:
AspectJ pointcut declarations are empty methods, should be ignored.


chrismair:
Thanks for the contribution!

PR #577, opened 2020-12-18
Ignore pointcut method params in UnusedMethodParameterRule

tlefevre:
Pointcuts can have method parameters that are not used

### DIFF
--- a/src/main/groovy/org/codenarc/rule/unused/UnusedMethodParameterRule.groovy
+++ b/src/main/groovy/org/codenarc/rule/unused/UnusedMethodParameterRule.groovy
@@ -43,9 +43,7 @@ class UnusedMethodParameterAstVisitor extends AbstractAstVisitor {
 
     @Override
     protected void visitConstructorOrMethod(MethodNode node, boolean isConstructor) {
-        if (!currentClassNode.isInterface() && !node.isAbstract() && !(currentClassName ==~ rule.ignoreClassRegex)
-            && !Modifier.isPrivate(node.modifiers) && AstUtil.getAnnotation(node, 'Override') == null
-            && !isMainMethod(node)) {
+        if (shouldEvaluateMethodParameters(node)) {
             def unusedParameterNames = node.parameters*.name
             def collector = new ReferenceCollector()
             collector.visitMethod(node)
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/codenarc/rule/unused/UnusedMethodParameterRule.groovy
```diff
@@ -43,9 +43,7 @@ class UnusedMethodParameterAstVisitor extends AbstractAstVisitor {
 
     @Override
     protected void visitConstructorOrMethod(MethodNode node, boolean isConstructor) {
-        if (!currentClassNode.isInterface() && !node.isAbstract() && !(currentClassName ==~ rule.ignoreClassRegex)
-            && !Modifier.isPrivate(node.modifiers) && AstUtil.getAnnotation(node, 'Override') == null
-            && !isMainMethod(node)) {
+        if (shouldEvaluateMethodParameters(node)) {
```
chrismair, line 46: I agree the code looks better and cleaner as you have refactored it, but it now performs all of the processing for all of the conditions, whether or not they are necessary, instead of short-circuiting, and some of that logic is non-trivial. At the risk of *premature optimization*, I'm inclined to leave as it was. What do you think? I would be fine with adding some well-named private helper methods, corresponding to those individual conditions, to improve readability and reduce the complexity/clutter.

tlefevre: Yeah, that makes sense - I'll rewrite it when I have the time.

chrismair: We are planning for the next CodeNarc release within the next couple weeks. Are you still intending on reworking this PR?

tlefevre: Ah sorry! It totally slipped my mind. I'll look at it soon :)

chrismair:
I manually added that change described above. Thanks.

tlefevre:
Ah! Sorry Chris, been really having trouble keeping up with work - thanks for the help!

Issue #578, opened 2020-12-30
Add ability for UnusedPrivateFieldRule to raise violations for fields which are only used within constructors


erdi:
If a field is only used within constructors then:
- if the usages are assignments then they are safe to be removed together with the field
- if the usages are not only assignments then the field should be changed into a local variable

PR #579, opened 2020-12-30
Add ability for UnusedPrivateFieldRule to raise violations for fields which are only used within constructors.

erdi:
PR for #578.

### DIFF
--- a/src/test/groovy/org/codenarc/rule/unused/AbstractUnusedPrivateFieldRuleTest.groovy
+++ b/src/test/groovy/org/codenarc/rule/unused/AbstractUnusedPrivateFieldRuleTest.groovy
@@ -1,0 +1,446 @@
+/*
+ * Copyright 2009 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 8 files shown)


src/test/groovy/org/codenarc/rule/unused/UnusedPrivateFieldRuleTest.groovy
```diff
@@ -15,416 +15,19 @@
...
+ */
+class UnusedPrivateFieldRuleTest extends AbstractUnusedPrivateFieldRuleTest {
 
     @Test
     void testRuleProperties() {
         assert rule.priority == 2
         assert rule.name == 'UnusedPrivateField'
     }
 
-    @Test
```
erdi, line 35: All these tests have been lifted to `AbstractUnusedPrivateFieldRuleTest`

src/test/groovy/org/codenarc/rule/unused/UnusedPrivateFieldRuleTest.groovy
```diff
@@ -15,416 +15,19 @@
...
-                    def myMethod(foo = ImportForm.this.importFilePath) {
-                    }
-                }
-            }
-'''
-        assertNoViolations(SOURCE)
-    }
-
     @Test
     void testBugFix_CannotCastFieldNodeToMetaClass() {
```
erdi, line 32: I left the test here because the usage is in a constructor

src/test/groovy/org/codenarc/rule/unused/AbstractUnusedPrivateFieldRuleTest.groovy
```diff
@@ -0,0 +1,446 @@
...
+    }
+
+    @Test
+    void testApplyTo_MultipleUnusedPrivateFields() {
+        final SOURCE = '''
+          class MyClass {
+              private int count
+              def otherField
+              private static String globalName = 'xxx'
+              void aMethod() {
```
erdi, line 57: I changed this test slightly - previously the code was using a constructor so I switched it to use a method so that the test passes regardless of constructor only usages being allowed or not.

chrismair:
Nice work! Thanks for the contribution.

Issue #582, opened 2020-12-30
Add NoBlankLineBeforeAnnotatedField rule


erdi:
It should raise violations when there is no blank line before a field declaration that uses annotations.

Example violations:
```
    class MyClass {
        // No violations for field declarations preceded by a comment
        @Delegate
        AutoCloseable stream
        
        String publicField                  // violation
        @PackageScope
        String packageScopedField
    }
```

chrismair:
I renamed `NoBlankLineBeforeAnnotatedField` rule to `MissingBlankLineBeforeAnnotatedField` to be consistent with the other "MissingXxx" rules in the same ruleset.

erdi:
Makes sense.

PR #583, opened 2020-12-30
Add NoBlankLineBeforeAnnotatedField rule.

erdi:
PR for #582.


erdi:
Merged `master` in to resolve conflicts and make it easier to accept and merge this PR.

chrismair:
Can you add a test for a field that has multiple annotations across more than one line.

erdi:
Sure thing, will do. 👍  I'm a bit busy at the moment but I should have time to do this on Sunday.

erdi:
> Can you add a test for a field that has multiple annotations across more than one line.

This has now been done. I did not know which case you actually meant so I added a test for when there is a violation and one where there isn't for fields annotated in such way.

chrismair:
Looks great. Thanks!

Issue #584, opened 2020-12-31
Add new MethodDeclarationNameTrailingWhitespace rule


erdi:
It should check whether method declarations do not contain unnecessary whitespace between method name and the opening parenthesis for parameter list.

Examples of violations:
```
    class ClassWithWhitespaceInConstructorDeclaration {
        
        ClassWithWhitespaceInConstructorDeclaration () { //violation
        }
        
        void methodWithWhitespaceInDeclaration () { //violation
        }
    }
```

PR #585, opened 2020-12-31
Add MethodDeclarationNameTrailingWhitespace rule.

erdi:
PR for #584.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ Build and Infrastructure
  - #570: Publish a fat jar. ([Eric Citaire](https://github.com/ericcitaire))
  - #573: Add Dockerfile to build CodeNarc Docker image using Gradle and Shadowjar.
 
+New Rules
+- #584: New **MethodDeclarationNameTrailingWhitespaceRule** rule (formatting) - Checks whether method declarations do not contain unnecessary whitespace between method name and the opening parenthesis for parameter list. ([Marcin Erdmann](https://github.com/erdi))
+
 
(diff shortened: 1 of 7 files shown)


erdi:
Merged `master` in to resolve conflicts and make it easier to accept and merge this PR.

chrismair:
That rule name is unwieldy, but I cannot think of a better one. *Naming is hard!* Thanks for the contribution.

chrismair:
I am wondering about the inconsistency of the name of this rule with the other `SpaceAfterNotOperator` rule, especially since this rule and `MethodCallNameTrailingWhitespace` are not really checking for "trailing" whitespace, but rather checking for space "in between", like `SpaceAfterNotOperator` does.

I am considering renaming these rules to `SpaceAfterMethodDeclarationName` and `SpaceAfterMethodCallName`. Any thoughts?

erdi:
Yeah, _naming is hard_ especially when naming things in what is not your first language so I'm always happy to accept ideas from others when it comes to naming. I think that your suggestions of `SpaceAfterMethodDeclarationName` and `SpaceAfterMethodCallName` are a) better and b) more consistent with `SpaceAfterMethodCallName`.

Would you like me to provide a PR with renaming these rules or would you like to rename them yourself? Keep in mind that it might be a bit more than just simply renaming the rules and their classes - I might have used the word "trailing" in descriptions and documentation...

chrismair:
If you are willing and able to provide a PR for the renaming of those two rules, that would be great. Thanks.

erdi:
Sure thing, will do.

PR #598, opened 2021-02-19
Fixed NPE, added some logging for when we encounter an invalid rule name

wheelerlaw:
If there is an invalid rule name passed as a part of the `ruleset` argument, a NPE will be thrown. This fixes it so it fails gracefully, reporting the issue, and moves on to the next rule.

### DIFF
--- a/src/main/groovy/org/codenarc/ruleset/JsonReaderRuleSet.groovy
+++ b/src/main/groovy/org/codenarc/ruleset/JsonReaderRuleSet.groovy
@@ -19,6 +19,8 @@ import groovy.json.JsonSlurper
 import org.codenarc.rule.Rule
 import org.codenarc.ruleregistry.RuleRegistryHolder
 import org.codenarc.util.PropertyUtil
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 
 /**
  * A <code>RuleSet</code> implementation that parses Rule definitions from Json read from a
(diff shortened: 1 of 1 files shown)


chrismair:
I agree that the NPE with no useful info needs to be fixed. The other ruleset implementations (Groovy DSL, XML) do not warn+ignore invalid rule names, but do throw an exception that includes the invalid rule name, e.g. “No such rule named [ClassName]”.

I would prefer to have the JSON ruleset behave the same way. So I'd like to change only the two lines starting on line 64, including adding the null-safe operator that you identified:
```
                RuleRegistryHolder.ruleRegistry?.getRuleClass(ruleName)?.name // ex: CatchThrowableRule
        assert ruleClassName, "No such rule named [$ruleName]"
```
and add this test to `JsonReaderRuleSetTest`:
```
    @Test
    void testRuleNameNotFound() {
        final JSON = '''
            {
                "DoesNotExist": {}
            }
            '''
        shouldFailWithMessageContaining('DoesNotExist') { parseJsonRuleSet(JSON) }
    }
```

Is that reasonable?

wheelerlaw:
Very. I will adjust the PR accordingly.

chrismair:
We are planning for the next CodeNarc release within the week or so. Are you still intending on reworking this PR?

chrismair:
I manually made the changes agreed to above.

PR #604, opened 2021-03-21
Adds ParameterAssignmentInFilter rule

mlkristiansen:
- Adds rule to avoid common error of assignment of variables or their properties in filtering closures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+
+#TODO: Sort the following line into the file
+- ParameterAssignmentInFilter rule (basic) - An assignment operator was used on a parameter in a filtering closure. This is usually a typo, and the comparison operator (==) was intended.
+
 <!-- markdownlint-disable MD003 MD004 MD007 MD032 -->
 # CodeNarc Change Log
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 
 #TODO: Sort the following line into the file
-- ParameterAssignmentInFilter rule (basic) - An assignment operator was used on a parameter in a filtering closure. This is usually a typo, and the comparison operator (==) was intended.
+- ParameterAssignmentInFilterClosure rule (basic) - An assignment operator was used on a parameter in a filtering closure. This is usually a typo, and the comparison operator (==) was intended.
 
 <!-- markdownlint-disable MD003 MD004 MD007 MD032 -->
 # CodeNarc Change Log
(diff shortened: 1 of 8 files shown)


mlkristiansen:
I am suggesting to add this rule to the list of rules in CodeNarc. Let me know if there is some procedure regarding suggesting rules I have missed.

chrismair:
Interesting idea. There is a significant overlap between this and the existing [ParameterReassignment](https://codenarc.org/codenarc-rules-convention.html#parameterreassignment-rule) rule, which also covers closures. I'm not sure that the extra bit this new rule adds warrants a new rule, along with all that overlap.

Would it be valuable to provide a simpler rule that checked for any assignment within those "filter" closures? Could you generalize and say that any assignment within those is bad practice and typically unintentional?

sbglasius:
I think the addition of this new rule makes sense, as it would prevent things like:
```
people.findAll { it.name = nameToFind } 
``` 
but that is not a `ParameterReassignment`

mlkristiansen:
@chrismair The assignment in filter was the initial idea for the rule, the problem is that, even within CodeNarc's own code, people introduced local vareiables in the filter closure. I don't think the local variables are a problem in that regard.

I have an alternate suggestions instead. Since parameter reassignment is already handled, I can change this to only take care of parameter property reassignment in filters like @sbglasius mentions. Is that a better idea?

chrismair:
Okay, on second thought, I have come around to your original proposal. The *ParameterReassignment* rule catches a "bad practice", but that is not a universal convention/rule and could be considered stylistic, while your rule catches a (likely) error. So I am fine with keeping the full behavior and checks you have here.

I am still undecided about the rule name. Do you like any of these better than *ParameterAssignmentInFilter*?
 - *ParameterAssignmentInFilterClosure*
 - *FilterMethodClosureParameterAssignment*
 - *ParameterAssignmentWithinFilterMethodClosure*

Naming things is hard (as always).

Some minor suggested changes:
 - Rather than recompute `FILTER_METHOD_NAMES.join('|')` for every method call, just make that a constant.
 - The markdown description for the new rule should list the (filter) methods that it is checking.
 - The markdown still lists the rule name a "AssignmentInFilter"; fix that to match the real name.

chrismair:
Are you planning on making any of the changes I suggested?

mlkristiansen:
Yes. i will go with the name ParameterAssignmentInFilterClosure and then implement the improvements.

mlkristiansen:
@chrismair I implemented the name change and your suggestions.

chrismair:
Nice. Thanks for the contribution.

PR #607, opened 2021-04-09
switched the violation in AssignCollectionSortRule

musketyr:
as the the violation contains reference to `$currentClassName`


chrismair:
Thanks for the fix!

PR #620, opened 2021-05-27
Fix #619 UnusedImportRule ignores AST imports

### LINKED ISSUE #619: UnusedImports triggers for AST introduced imports with no use
When Grails is introducing unused imports through AST transformations, these are reported without a line number.

image

sbglasius:
Fix issue


chrismair:
Thanks for the fix!

Issue #632, opened 2021-06-14
MissingBlankLineBeforeAnnotatedField false positive at the class start when class is annotated too 


dmurat:
MissingBlankLineBeforeAnnotatedField reports violation for the annotated field at the beginning of a class when the class itself is annotated.
The following example produces violation:
```
@SomeClassAnnotation
abstract class JerseySpec extends Specification {
    @Delegate
    private JerseyTest jerseyTest
}
```

```
Violation[rule=MissingBlankLineBeforeAnnotatedFieldRule[name=MissingBlankLineBeforeAnnotatedField, priority=3], lineNumber=5, sourceLine=private JerseyTest jerseyTest, message=There is no blank line before the declaration for field "jerseyTest" that has annotations]
```

dmurat:
Here is a bug fix proposal: https://github.com/CodeNarc/CodeNarc/pull/633

PR #633, opened 2021-06-14
MissingBlankLineBeforeAnnotatedFieldRule should work when class is annotated

dmurat:
Proposed bug fix for https://github.com/CodeNarc/CodeNarc/issues/632


chrismair:
Thanks for the fix!

Issue #611, opened 2021-04-18
BracesForMethod - allow braces in new line for multiline declarations


dmurat:
After upgrading to Codenard 2.1.0, BracesForMethod starts failing for me in some cases. I guess this is because of fixes in #603.

When having methods/constructors with many parameters, I'm placing curly braces in a new line like in the following example:
```
protected HttpResponseMetaData localizeHttpResponseMetaData(
    HttpResponseMetaData httpResponseMetaData, HandlerMethod handlerMethod, Locale locale, ConstraintViolationException constraintViolationException)
{
  ...
} 
```

Is it possible to add a new option to the BracesForMethod to support such formatting? Something like `allowNewLineForMultilineDeclarations` which should be taken into account only if the `sameLine` is `true`.

Tnx

chrismair:
Hmmm, you put the opening brace on the same line *unless* it is multiple lines like that? That seems like a very special-case customization. I don't think I have encountered that style before, but I certainly don't claim to be familiar with all coding styles.

I encourage others to chime in and vote for that if it has broader support.

dmurat:
If it can help my case :-), this kind of formatting is supported in IDEA (and Eclipse I believe) for years.

erdi:
FWIW I haven't encountered such formatting style either. I find it unusual that the opening brace is on the new line in this case but not when the declaration does not span multiple lines. I'm personally more used to formatting this case like this:
```
protected HttpResponseMetaData localizeHttpResponseMetaData(
    HttpResponseMetaData httpResponseMetaData, HandlerMethod handlerMethod, Locale locale, 
    ConstraintViolationException constraintViolationException
) {
  ...
} 
```

dmurat:
Of course, there is nothing wrong with any style. Both are a personal preference, I think. 

However, I will like to point that "my style" is not really something new. It is quite well supported in IDEs and other Java tools. It is just a multiline extension of a style that always puts a brace in the new line. I believe that most of the C# code is formatted like this.

If I try to reason about it, I would say that the point of "my style" is the separation of declaration and the code block that follows behind it. I don't really need to see a bracket to know that parameter declaration is over, just need a brace for opening a code block. And there is some additional order in having opening and closing braces to be on the same column.

As speaking about personal preference, I don't want to persuade anybody to use it, but don't see any harm to have it supported by Codenarc. Otherwise, I will miss it for sure, as anybody else would miss their own preferred style, I guess.

Tnx.

[...]

PR #635, opened 2021-06-15
BracesForMethod - allow braces in new line for multiline declarations

dmurat:
Here is a proposal for BracesForMethod rule enhancement as described in https://github.com/CodeNarc/CodeNarc/issues/611


chrismair:
Thanks for the contribution. Two suggestions:
 - How about renaming the *whenSameLineAllowNewLineForMultilineDeclarations* property to *allowBraceOnNextLineForMultilineDeclarations*?
 - What if the next line contains the brace followed by a comment? Would it be better and sufficient to check if that next line starts with '{'? Add test for that scenario.

dmurat:
I updated the code as requested. If there is anything else I can do, please let me know. Thank you.

chrismair:
Looks good. Thanks!

Issue #636, opened 2021-06-17
"CodeNarc." prefix does not work for method level annotations


dmurat:
I like the possibility to prefix rule names in @SuppressWarnings annotations (introduced by #610). Unfortunately, it only works for class-level suppressions, and not for method-level suppressions.
I believe the problem is in `org.codenarc.analyzer.SuppressionAnalyzer.isViolationSuppressed(Violation violation)` method where `CodeNarc.` prefix is not taken into account:

```
    public boolean isViolationSuppressed(Violation violation) {
        ...
        BitSet lines = suppressionsByLineNumber.get(ruleName);
        if (lines != null) {
            return lines.get(lineNumber);
        }

        // this part is missing, I believe - start
        lines = suppressionsByLineNumber.get(CODE_NARC + "." + ruleName);
        if (lines != null) {
            return lines.get(lineNumber);
        }
        // this part is missing, I believe - end

        return false;
    }
```

However, I'm not quite sure if the intention was to use the feature in this way. If not, please reconsider. I think it is very useful.

Tnx

dmurat:
@musketyr Can you, please, comment on this. Tnx.

musketyr:
Thank you for detailed analysis. The intention was Indeed to work in the other scopes as well not only on the class level.  I'll try to take a look tomorrow.

musketyr:
see the attached PR #637

PR #637, opened 2021-06-21
Made CodeNarc. prefix suppression work for fields and methods

musketyr:
fíixes #636


chrismair:
Thanks for the fix!

musketyr:
Sure thing!

PR #640, opened 2021-07-11
Fix parameter table for PackageName Rule documentation

goober:
The parameter table for the `PackageNameRule` documentation has some extra cells, this pull request fixes the formatting.


chrismair:
Thanks for the fix!

Issue #662, opened 2022-01-20
MissingBlankLineBeforeAnnotatedFieldRule reports invalid violation when class annotation with line comment is present


dmurat:
After adding the following test in `MissingBlankLineBeforeAnnotatedFieldRuleTest`, I'm getting violations reported for `jerseyTest`:

```
    @Test
    void test_AnnotationOnTheFirstLineOfTheClassWithAnnotationAndLineComment_NoViolation() {
        final SOURCE = '''
            @SomeClassAnnotation3 // Some comment
            abstract class JerseySpec extends Specification {
                @Delegate
                private JerseyTest jerseyTest
            }
        '''
        assertNoViolations(SOURCE)
    }
```

Tnx

dmurat:
Here is proposed PR: https://github.com/CodeNarc/CodeNarc/pull/663

PR #663, opened 2022-01-20
MissingBlankLineBeforeAnnotatedFieldRule proposed fix

dmurat:
This is a proposed PR for https://github.com/CodeNarc/CodeNarc/issues/662

The solution proposed relies on changes in `AstUtil` which might have wide consequences.
I didn't find problems, except with `IndentationRuleTest` that requires a small change.
However, it looks good to me because when reporting violations for annotated node, `AbstractAstVisitor.addViolation(ASTNode, String)` reports the first non-annotation line as a violation.


chrismair:
Thanks for the fix. I may end up reworking (hopefully simplifying) that AstUtil logic at some point. Moving to Groovy 3 means we can jettison some of the older hacks/workarounds necessary for earlier Goovy versions.

Issue #282, opened 2017-12-28
Add support for `ignore` list


tmysik:
For existing project, if one wants to start using CodeNarc, it is quite difficult because usually, there is a lot of existing violations against the rule sets and it is not easily possible to fix them first (about ~50k violations of all priorities in our case). So, in such a case, it would be very useful to be able to add these existing violations to some ignore list and report only **new** violations. The steps then would be quite simple:

* run ConeNarc for the 1st time with XML report
* save this report as our initial ignore list
* run CodeNarc again with our ignore list (e.g. `-ignorefile=<the-original-xml-report>`)

-> only new violations would be reported.

Thanks!

erdi:
Isn't http://codenarc.sourceforge.net/codenarc-BaselineXmlReportWriter.html what you're looking for?

tmysik:
I see, thanks! Yes, this seems to be like what I wanted but - this cannot be used from command line directly, right? Only with Ant? Or am I missing something? Thanks a lot.

chrismair:
Correct, that is currently only supported on the Ant task.

tmysik:
Thank you. So, I will leave this ticket open, I can imagine that it would be useful also for the CLI interface.

chrismair:
That seems reasonable.

tmysik:
One note to the *baseline* functionality: I generated baseline file for our project (about 16 MB ;) but if I use it later for a test run, I still get some violations even if there should be none. I needed to disable 2 specific rules [1] to be able to pass the test. I know that it is beta quality now but still - should I report it?

Thank you.
[1]
```properties
DuplicateStringLiteral.enabled=false
UnnecessaryGString.enabled=false
```

chrismair:
Yes, please do create an issue for that. But we'll probably need more details. That facility uses some heuristics to try to match up original violations with new ones, and that may not always be sophisticated enough, and may produce some _false negatives_ or _false positives_.

Did the code change between the original generation of that baseline file and the new CodeNarc run, especially for the file(s) where the unexpected violations occurred?

tmysik:
@chrismair No, the source code did not change. Reported as issue #303, hope it helps.

Thanks!

chrismair:
#303 was fixed and closed.

UlrichEckhardt:
[...]

PR #673, opened 2022-03-08
CodeNarc: Implement a CLI for baseline usage.

UlrichEckhardt:
This complements the setting in the Ant task, where you can define a baseline. You specify the baseline using `-excludeBaseline=<FILENAME>`. It is aimed to solve #282.

This branch is based on v2.2.0, even though the commit cleanly applies to master as well. The problem is that I can't get the tests to work there, even without my changes. Still, this is something to try out and review.


chrismair:
From my initial review, this looks like the right approach. Updating the associated doc in "codenarc-command-line.md" would also be appreciated. Thanks.

UlrichEckhardt:
Just updated the pull request, found some other places, too.

Just a side question: The instructions in docker/README.md, they pull a specific version of CodeNarc from GitHub and build an image from that, right? Is there a way to create one from the local sources?

UlrichEckhardt:
## Docker Image Build

Okay, managed to build a Docker image functioning equivalently to the the one from dockerhub:
```Dockerfile
#### Build Image
FROM gradle:6.9-jdk11-alpine AS gradle

WORKDIR /gradle
COPY . .
RUN gradle --quiet --no-daemon shadowJar

#### Runtime Image
FROM openjdk:11-jre-slim

COPY --from=gradle /gradle/build/libs/CodeNarc-2.2.0-all.jar /lib/codenarc-all.jar
RUN chmod 555 /lib/codenarc-all.jar

WORKDIR /ws

ENTRYPOINT ["java", "-jar", "/lib/codenarc-all.jar"]
```

## Pull Request

Using the Docker image, I was able to verify that creating and using baseline files works from the commandline, making this pull request ready for review. I ran the tests using Gradle inside of a Docker container as well, but that already worked before.

## Future Ideas

From using other linters with baselines, I can say that there are a few things missing to make this really useful:
 * Fail on unexpected error: When encountering any error not in the baseline, the linter should signal an error. Yes, you can use the three `-maxPriority*Violations=0` options, but that's cumbersome and not very elegant.
 * Fail on unexpected success: When not encountering an error that is in the baseline, the linter should signal an error as well. This forces the use to regenerate the baseline and/or manually edit it. However, it prevents the error from being re-introduced with a second change.
 * Timestamp: I wonder if it's necessary to include it (`timestamp="Mar 10, 2022, 7:12:12 PM"`) in the baseline, even more so since it's using a localized, timezone-dependent format. This simply makes builds (and linting is a kind-of build) non-repeatable. More practically, I couldn't implement the above "fail on missing error" feature myself by just running CodeNarc and comparing with the previous results, because every run differs.

chrismair:
Thanks for the contribution!

chrismair:
And thanks for going through the extra work required to test it properly.

UlrichEckhardt:
Thank you for your work as well, Chris! Do you have any release plans? I actually have a case where I want to use this feature in a Jenkins pipeline and I would rather not use an unreleased version there.

chrismair:
I'm targeting a release within the next several days.

PR #685, opened 2022-04-02
Fix running tests

darxriggs:
Without enabling the JUnit feature no tests are run when using `./gradlew check`.

Reference: https://junit.org/junit5/docs/5.8.2/user-guide/#running-tests-build-gradle


chrismair:
Thanks!

PR #687, opened 2022-04-02
SpaceAfterMethodCallName: Fix handling methods with special characters

### LINKED ISSUE #686: SpaceAfterMethodCallName fails for methods with special characters
When using [withFormat](https://docs.grails.org/5.1.x/ref/Controllers/withFormat.html) in Grails in combination with a wildcard `*` as follows:

```
withFormat {
    json { ... }
    '*' { ... }
}
```

...the `SpaceAfterMethodCallNameRule` fails with this issue:

```
Error from [org.codenarc.rule.formatting.SpaceAfterMethodCallNameRule] processing source file [null]
java.util.regex.PatternSyntaxException: Dangling meta character '*' near index 0
*\s+\(
```

darxriggs:
This fixes #686.


chrismair:
Thanks for the fix René !

Issue #684, opened 2022-03-28
Baseline improvements


UlrichEckhardt:
I'm looking at the baseline support in the context of an existing project currently and have found a few things that feel wrong to me at the moment:

 * The timestamp in the header changes on every run.
 * The ordering of XML "File" blocks seems to be random. I guess it uses the ordering that just happens to be in the filesystem.

That means that you can't easily compare the results of two runs, i.e. the results are not deterministic.

Another related issue is that unused exceptions (the baseline is effectively a set of exceptions to the configured rules) are not signalled. If I fix an issue, there is nothing prevents it from coming back. Also, when I finally regenerate the baseline, I get a bunch of unrelated changes in the file, making it hard to review. As a workaround, I wanted to simply regenerate the baseline and run `git diff --exit-code` on it, but that doesn't work since the baseline isn't deterministic.

I'd tackle this in three steps:

 * Remove the timestamp from the baseline.
 * Sort the "File" blocks alphabetically, preferably using an ordering that does not depend on e.g. the language settings.
 * Extend the baseline handling to signal unused entries as error.

Thoughts? Ideas? Suggestions?

chrismair:
The first two items sound fine. Good ideas.

> Extend the baseline handling to signal unused entries as error.

I'm interested in how you would implement that. And we'd want to try hard to maintain _backward-compatibility_. I would not want to break everyone's existing baseline files, and if that was unavoidable, then those changes would have to wait for the next major release (which is not going to be anytime soon).

UlrichEckhardt:
I'm thinking of using something similar to the decorator pattern:
 * The engine itself emits a stream of issues.
 * Every issue passes the decorator. If an issue is matched by a baseline entry, both the entry and the issue are discarded. Unmatched issues are forwarded.
 * After all issues from the engine were processed, any remaining entries in the baseline cause an "unused entry" issue to be emitted.

I think the algorithm is relatively simple to implement, it just takes work.

chrismair:
Okay that sounds promising. Please include testing with an old/existing baseline file to make sure that still works.

UlrichEckhardt:
Finally got to it on the weekend, please take a look at #689 for a first step.

chrismair:
Feel free to reopen if still needed.

PR #689, opened 2022-04-11
Feature/deterministic baseline report

UlrichEckhardt:
Two changes here:

- Don't write a timestamp.
- Sort entries in the baseline by their file path.

This should be a first step towards resolving #684.


src/test/groovy/org/codenarc/report/BaselineXmlReportWriterTest.groovy
```diff
@@ -32,7 +32,7 @@ class BaselineXmlReportWriterTest extends AbstractXmlReportWriterTestCase {
```
chrismair, line 32: Can you reorder and /or add to the Directory/File Results so that the test exercises the new sorting?

chrismair: Changes look good. Thanks.

The _numberOfFilesInThisDirectory_ is used in `getTotalNumberOfFiles()`. You are more than welcome to explore a simpler safe+clean way to preserve that functionality in a separate PR. Better code is always welcome.

PR #694, opened 2022-04-18
Feature/report test refactoring

UlrichEckhardt:
As already mentioned in https://github.com/CodeNarc/CodeNarc/pull/689#discussion_r850163853, I've been looking at the `numberOfFilesInThisDirectory` variable in class DirectoryResults. Before tackling that part though, here are some cleanups related to the tests. The idea was to first make the tests correct (i.e. ensure that the counter is set to two if there are two files) and then to replace the counter, because it is much easier to review those things separately. There were quite a few places in tests where the counter wasn't set correctly or maintained and in addition that influenced quite some of the test results. This PR deals with those things first.


chrismair:
I'll have to review/merge this next week. The initial review looks good.

chrismair:
Thanks for that cleanup!

PR #695, opened 2022-04-18
Refactoring directory handling in results

UlrichEckhardt:
This one is based on #694 and already presented as a draft here. There are two main elements to this, maintaining a file count in `DirectoryResults` and handling of files without violations. Actually, I could even split this PR at their border, if desirable.

The handling of the file count in `DirectoryResults` is the easier part of the two. Instead of having a member variable, just look at the list of children. The rest is just to remove code that initialized or modified the count. Actually, the hard work was sorting the cases where this changed the results for the tests into those where the tests themselves where wrong and cases where the code was actually wrong.

The other part is the handling of files without violations. There was some special code inside the two classes `FilesystemSourceAnalyzer` and `AntFileSetSourceAnalyzer` that stripped files without violations from the results, but only partially. I've removed that code, so that even files without violations are included now. If and where a report doesn't want is now implemented there.

### DIFF
--- a/src/main/groovy/org/codenarc/analyzer/FilesystemSourceAnalyzer.groovy
+++ b/src/main/groovy/org/codenarc/analyzer/FilesystemSourceAnalyzer.groovy
@@ -131,7 +131,6 @@ class FilesystemSourceAnalyzer extends AbstractSourceAnalyzer {
         if (matches(sourceFile)) {
             List allViolations = collectViolations(sourceFile, ruleSet)
             def fileResults = new FileResults(filePath, allViolations, sourceFile)
-            dirResults.numberOfFilesInThisDirectory++
             dirResults.addChild(fileResults)
         }
     }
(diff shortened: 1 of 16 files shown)


UlrichEckhardt:
Heya @chrismair,

stumbled over this PR which I prepared but never completely finished, since I stopped having anything to do with Jenkins and Groovy at that time. Anyhow, I just rebased it, adjusted a bit of code that had been changed since then and now it's ready for review.

Just to elaborate on the motivation behind this, it was to simplify things basically. The point is that storing the number of files in a directory and the actual children (subdirs and files) separately means that there is redundant information, which must carefully be kept in sync. Just counting the children when they are needed isn't a big deal and makes the separate counter redundant, leading to less code overall.

Greetings from Hamburg!

Uli

chrismair (approved):

chrismair:
Thanks very much, Uli. I like that result a lot better!

PR #699, opened 2022-04-21
UnusedImportRule: detect property-style usage of static imports

### LINKED ISSUE #698: UnusedImport wrongfully reports unused imports when using property-syntax
```groovy
import static System.getProperties
class ABC {
    def foo() {
        properties.getProperty("foo")
    }
}
```

This should not report a violation.

leonard84:
Prior to this commit, UnusedImportRule didn't detect usage of static imports of getter/setter-style methods
when written as properties.

fixes #698


chrismair:
Thanks for the fix/enhancement!

PR #707, opened 2022-05-04
CodeNarc#706: Allow functioning with Groovy 4

paulk-asert:


chrismair:
These changes were reverted, in favor of publishing a Groovy 4-specific version of CodeNarc. See #709

PR #708, opened 2022-05-27
Feature/handling files without violations

UlrichEckhardt:
This makes sure that at the analyzer level, files are treated regardless of whether they contain violations or not. Instead, when generating a report from the analyzer results, skip those files.

Also, add a file without violations to most tests, just to ensure correct behaviour. BTW: Here's one thing that could be worth refactoring: All different report generator tests have their own set of analyzer results they operate on. Replacing that with a centrally managed set of analyzer results would make test maintenance a bit more convenient. Also, if you add example data for _one_ corner case, you'd know that you have it covered for _all_ report generators.

I extracted this part from #695, which was too big to realistically review.

### DIFF
--- a/src/main/groovy/org/codenarc/analyzer/FilesystemSourceAnalyzer.groovy
+++ b/src/main/groovy/org/codenarc/analyzer/FilesystemSourceAnalyzer.groovy
@@ -97,7 +97,7 @@ class FilesystemSourceAnalyzer extends AbstractSourceAnalyzer {
     @SuppressWarnings(['CatchThrowable', 'NestedBlockDepth'])
     private DirectoryResults processDirectory(String dir, RuleSet ruleSet) {
         def dirResults = new DirectoryResults(dir)
-        def dirFile = new File((String) baseDirectory, (String) dir)
+        def dirFile = new File(baseDirectory, dir)
         dirFile.eachFile { file ->
             def dirPrefix = dir ? dir + SEP : dir
             def filePath = dirPrefix + file.name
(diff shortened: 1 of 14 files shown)


chrismair:
Thanks Ulrich. I appreciate you taking the time to do this.

I'm a little hesitant to just jump on a redesign like this. What is the necessity/benefit of going from a (intentionally) sparse graph of files-with-violations to a full graph of every source file (which could be thousands or more)?

UlrichEckhardt:
Thank you for questioning this, Chris! Let me try to explain my thoughts...

First off, I don't think that having a simple data structure for a file without violations is much overhead. Reading, parsing and scanning the files probably dwarfs this. With that in mind, there is the question where the "empty" files should be filtered. Here, my thought is that it's in the responsibility of whatever generates the report. Whether you want to include a paragraph on files without violations isn't something I would decide in the place where I'm doing the scanning. I do see the advantage of skipping them there though, it's a simple optimization with negligible risk and I might have done the same.

The place where this was getting in my way was when I tried to improve the baseline handling. There, I really want the full set of files to distinguish between a file that wasn't scanned and a file that was scanned but doesn't contain any violation. If the baseline contains an exception for this file, that would mean the baseline needs to be adjusted after fixing the file. It would be hard to reconstruct that info there.

Also, there is a `numberOfFilesInThisDirectory` counter in DirectoryResults, which sticks out to me. On first look, it seems to be redundant, because a simple recursive scan of `children` should provide the same info. On closer look, it doesn't always represent the result you would get from `children`, which may be a feature or a bug. In a follow-up set of commits, I do away with that, too, so that this surprising behaviour is gone.

chrismair:
Yes, I agree that the extra processing time overhead is likely negligible. I expect it will have a  bigger impact on memory usage for large projects. But even that is probably acceptable -- though I have experienced my share of pain from unexpected consequences after changes like that.

It will be a while before I can focus on this. I'm working on Groovy 4.x compatibility, and that likely includes reverting some other changes.

UlrichEckhardt:
Take your time! I guess you get paid as much as I do, meaning we both do it for fun and fame, so that's totally okay. :)

I'll keep working on the follow-up parts of my overall goal of getting the baseline support in the shape I want it to have. This PR can remain lingering here, I can deal with that.

Have a good one!

chrismair:
Okay, I finally got around to fully reviewing the code changes, and I also pulled them in locally and tested against several large Groovy codebases to make sure the results remained consistent before and after. Looks good. Thanks for the contribution!

PR #715, opened 2022-08-18
SpaceBeforeOpeningBraceRule and SpaceAfterOpeningBraceRule fail when …

leonard84:
higher unicode characters are used on the same line.

This PR provides a reproducer for this issue.


chrismair:
Thanks for providing the failing test cases.

chrismair:
Will be available with CodeNarc v3.2.0

PR #719, opened 2022-11-01
Remove CodeNarc Eclipse plugin reference

darxriggs:
The plugin has been discontinued and removed from the Eclipse Marketplace on 2022-11-01.


chrismair:
Thanks René!

Issue #712, opened 2022-08-05
Label indent - question 


dkubicki:
Hi team, 

I'm looking for an option to check if the label indent level is correct.
Especially for Spock's label:

Failed

```    
       def "should sth"() {
        given:
        RevokeRequest request = aRevokeRequest()
```
Passed 

```    
       def "should sth"() {
        given:
             RevokeRequest request = aRevokeRequest()
```
Do we have such an option in the current set rules?

chrismair:
CodeNarc does have an [Indentation](https://codenarc.org/codenarc-rules-formatting.html#indentation-rule) rule, but it does not check labels. And it looks like the labels are not exposed by the Groovy AST, so it is likely not practical to add such checks.

xmac11:
Hello @chrismair @dkubicki , I came across the same issue which prevents using the `Indenation` rule with Spock, since the following syntax is considered a violation:

```groovy
def "should be true"() {
    expect:
        1 == 1
```

I've opened a PR addressing this (https://github.com/CodeNarc/CodeNarc/pull/729) 
Please have a look if it's something that you would want CodeNarc to support, otherwise feel free to close it.

PR #729, opened 2023-03-05
Enhance IndentationRule to take into account Label Indent

xmac11:
Fixes https://github.com/CodeNarc/CodeNarc/issues/712

### Description
Currently, the following syntax is considered as a violation of `IndentationRule`

```groovy
def "should be true"() {
    expect:
        1 == 1
```

This PR introduces a `indentUnderLabel` boolean property to make this configurable. 
The default value is `false` in order to be backwards compatible.
When it is set to `true`, then code within the label should be indented with the number of spaces defined by the `spacesPerIndentLevel` property.

Additionally, the following change is made in order to allow labels without a text description.

```diff
private static boolean isSpockBlockLabel(Statement statement) {
    return (statement.statementLabel in SPOCK_BLOCKS &&
            statement instanceof ExpressionStatement &&
-           statement.expression.class == ConstantExpression &&
-           statement.expression.type.clazz == String
    )
}
```

### How has this been tested
Unit test are added covering:
- No violations, including multiple nested blocks
- No violations for labels with and without a description
- Violations when `indentUnderLabel` property is not respected

### DIFF
--- a/src/main/groovy/org/codenarc/rule/formatting/IndentationRule.groovy
+++ b/src/main/groovy/org/codenarc/rule/formatting/IndentationRule.groovy
@@ -40,6 +40,7 @@ class IndentationRule extends AbstractAstVisitorRule {
     int priority = 3
     Class astVisitorClass = IndentationAstVisitor
     int spacesPerIndentLevel = 4
+    boolean indentUnderLabel = false
 
     // Global Map of indent levels by class; enables coordination across different Class Nodes.
     protected final Map<SourceCode, Map<ClassNode, Integer>> classNodeIndentLevels = new ConcurrentHashMap<>()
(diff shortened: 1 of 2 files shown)


chrismair:
Nice contribution. I like it.

Would it be acceptable and sufficient to just have a boolean that controls whether the statements under a label should be indented a single level or not?

The existing rule configuration property is `spacesPerIndentLevel`, and the rest of the rule logic is concerned with the *indent level*. Could we have this new behavior align with that? I'm hoping it is not a requirement that the statements under a label be indented for some number of spaces other than `spacesPerIndentLevel`.

xmac11:
Good point @chrismair, I was also wondering the same. 
The only reason I went with this approach is that Groovy Code Style exposes a separate property for `Label indent`.

Screenshot 2023-03-08 at 10 01 00 AM

But I suppose most users will be using the same number of spaces everywhere.
Please let me know what you prefer and I'll go with it.

xmac11:
@chrismair done :)

chrismair:
Looks good. Thanks for the contribution.

Issue #739, opened 2023-08-04
NoScriptBindings: False positives for variables re-assigned in if block


stefan6419846:
Suppose I have the following code snippet on the top level without a corresponding class (which is a simplified version of more complex conditional code I am linting):

```groovy
String getValue(boolean isActive) {
    String value = 'abc'
    if (isActive) {
        value = 'def'
    }
    return value
}
```

Running `org.codenarc:CodeNarc:3.3.0-groovy-4.0` with most of the rules enabled on this code yields:

> Violation: Rule=NoScriptBindings P=3 Line=4 Msg=[The script variable [value] does not have a type declaration. It will be bound to the script which could cause concurrency issues.] Src=[value = 'def']

This seems like a false positive for me, as the variable *value* should indeed have method scope. Re-declaring *value* inside the if block will result in an *UnusedVariable* warning in CodeNarc and a syntax error on the file itself, thus not being the correct solution either:

> The current scope already contains a variable of the name value

chrismair:
Yes, that does look like a false positive and a bug in that new rule.

boris-github:
It is a false positive. Here is my example which triggers the [NoScriptBindings](https://codenarc.org/codenarc-rules-concurrency.html#noscriptbindings-rule) rule violation:

`The script variable [params] does not have a type declaration. It will be bound to the script which could cause concurrency issues.`

```
private myFunction(configFile) {
  def params
  if (fileExists(configFile)) {
    params = utils.readYamlFile(configFile)
  }
  params
}
```

FYI @jchorl

stefan6419846:
I just had some more thoughts about this, not having a strong Groovy background, but some experience with Python ASTs, including using them for linting. The reason is that I have quite some code where I am currently silencing these violations for complete sections due to the false positives.

AFAIK this is roughly similar to checking whether a variable has been defined in the current scope before with the assignment condition being replaced. Skimming through the existing code, CodeNarc does not have a corresponding rule, only the *UnusedVariable* one with its known limitation of not correctly respecting scopes.

chrismair:
This is a new rule in the latest release, and that is a pretty significant problem, so I would lean toward disabling the rule until that gets fixed, likely in the next release (probably in 4-6 months). 

If you are proposing another new rule, that deserves opening a separate issue. A pointer to existing rules in other languages is always helpful, and pull requests are always welcome.

PR #742, opened 2023-08-17
#739 Update NoScriptBindingsRuleTest.groovy

boris-github:
Added a scenario which should not result in the rule violation.


chrismair:
Thanks for the contribution. I ended up adding some separate test cases for this (reassigning variable) and also for reassigning parameters, which is also a problem.
[#739: Add test cases for known issues with NoScriptBindings rule, rea… · CodeNarc/CodeNarc@6e47fdf](https://github.com/CodeNarc/CodeNarc/commit/6e47fdfce109b5c4345011b82290a3c6ff84064f)

PR #753, opened 2023-11-07
fix: exact path patterns

### LINKED ISSUE #752: Impossible to match files in a specific directory
The following arguments don't match as intended:
```text
-includes=test3/*.groovy
-basedir=/home/myuser
```

Files:
```text
/home/myusers/test3/wibble.groovy
```

The combination of the base director + the include precisely matches the file listed above but CodeNarc fails to find it.

I've tracked this down to https://github.com/CodeNarc/CodeNarc/blob/master/src/main/groovy/org/codenarc/analyzer/FilesystemSourceAnalyzer.groovy#L125

Here the fully qualified path is passed into the pattern for matching, meaning it will never match.

The default pattern works because it includes the `**` so matching any sub directory, so changing the pattern to `**test3/*.groovy` was my first thought, however that means it would match anything that ends in `test3/*.groovy` not just the files in the `test3` sub directory of the user's home directory, for example it would match both these files:
```text
/home/myusers/test3/test4.groovy
/home/myusers/test2/test3/test5.groovy
```

As another example this also fails to match any files when run from `/home/myuser`
```text
-basedir=test3
-includes='wibble.groovy'
```

I believe the fix is to have the pattern operate on the path relative to the `basedir`.

One final confusion is why the default patterns are `**/*.groovy` as the functions document `**` as matching any file / directory so I believe if the bug above is fixed, the default pattern should only require `**.groovy`?

stevenh:
Fix processing of exact path patterns by matching the pattern against the file path (relative path), not the absolute path with basedir prepended (absolute path)

This allows users to specify exact paths within patterns, for example:
```text
-include=subdir/*.groovy
-basedir=/home/myuser
```

Will now correctly match the file:
```text
/home/myuser/subdir/*.groovy
```
This changes the behaviour of pattern matches to ignore a forward slash directly after a double star, for example:` **/*.groovy` so that it will match files in the directory or sub directories instead of just sub directories, matching the behaviour of ant / globstar.

Also:
* Add bin/ to .gitignore so output files aren't available for commit.
* Add helpful message on assert failure of equal sets.

Fixes #752

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ src/main/java/META-INF/
 /classes
 /out
 /integration-test/out/test/classes
+bin/
--- a/src/main/groovy/org/codenarc/analyzer/FilesystemSourceAnalyzer.groovy
+++ b/src/main/groovy/org/codenarc/analyzer/FilesystemSourceAnalyzer.groovy
@@ -125,7 +125,8 @@ class FilesystemSourceAnalyzer extends AbstractSourceAnalyzer {
     private void processFile(String filePath, DirectoryResults dirResults, RuleSet ruleSet) {
         def file = new File(baseDirectory, filePath)
         def sourceFile = new SourceFile(file)
-        if (matches(sourceFile)) {
+        // Match on the filePath only to allow for exact relative paths.
+        if (matches(new SourceFile(new File(filePath)))) {
             List allViolations = collectViolations(sourceFile, ruleSet)
             def fileResults = new FileResults(filePath, allViolations, sourceFile)
             dirResults.numberOfFilesInThisDirectory++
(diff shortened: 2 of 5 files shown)


chrismair:
Thanks very much for the contribution and working out that fix.

But I do not want to break existing users who may depend on the current behavior.

I created a separate branch/PR with a different fix that expands the matching so that it matches includes and excludes against the full source file path and ALSO against path relative to the baseDirectory.  I did use your test as-is, though. That was helpful, too.

[Feature/exact path patterns by chrismair · Pull Request #754 · CodeNarc/CodeNarc](https://github.com/CodeNarc/CodeNarc/pull/754)

stevenh:
> Thanks very much for the contribution and working out that fix.
> 
> But I do not want to break existing users who may depend on the current behavior.
> 
> I created a separate branch/PR with a different fix that expands the matching so that it matches includes and excludes against the full source file path and ALSO against path relative to the baseDirectory. I did use your test as-is, though. That was helpful, too.
> 
> [Feature/exact path patterns by chrismair · Pull Request #754 · CodeNarc/CodeNarc](https://github.com/CodeNarc/CodeNarc/pull/754)

Thanks I'll try test that to ensure it doesn't overmatch in the exact case.

chrismair:
Closing; replaced by #754

PR #754, opened 2023-11-12
Feature/exact path patterns

### LINKED ISSUE #752: Impossible to match files in a specific directory

chrismair:
Match includes and excludes against full source file path and also against path relative to the baseDirectory. 

Fixes #752

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ NOTE: CodeNarc now provides a separate **3.x.x-groovy-4.x** version for use with
 TODO: Version 3.?.?    (??? 202?)
 --------------------------------------
  - #744: Fix wording and typos. ([René Scheibe](https://github.com/darxriggs))
+ - #754: Command-line (FilesystemSourceAnalyzer): Match includes and excludes against full source file path and also against path relative to the `baseDirectory`.
 
 
 Version 3.3.0    (Jul 2023)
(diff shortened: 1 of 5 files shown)


stevenh:
Thanks for getting this done @chrismair any chance you could cut a release with this in so we can get [npm-groovy-lint](https://github.com/nvuillam/npm-groovy-lint/) updated?

stevenh:
Actually there's [still a related issue](https://github.com/CodeNarc/CodeNarc/issues/755), which needs to be fixed before npm-groovy-lint tests would pass.

PR #756, opened 2023-11-25
fix: relative paths on windows

stevenh:
Fix relative paths on Windows which didn't match due non normalized comparison with base directory.


chrismair:
Thanks for the fix!

PR #763, opened 2024-02-15
add new jenkins ruleset for code in Jenkinsfiles and Jenkins Pipeline Libraries

danielzaenker:
[Jenkins](https://www.jenkins.io/) Groovy code is executed using a special runtime that can produce unexpected behaviour called CPS Method mismatches. See this [article](https://www.jenkins.io/doc/book/pipeline/cps-method-mismatches/) for more information. Furthermore Jenkins requires you to almost always use Serializable Types which also can lead to runtime errors. Because it is very painful to debug these errors at runtime, we created these rules to detect them statically. Documentation about these problems is quite sparse, so a lot of this code is based on our experience and testing.

These make only sense when used in Jenkins code so we created a new "Jenkins" Ruleset.
The following rules were added:
* ClassNotSerializable
* ClosureInGString
* CpsCallFromNonCpsMethod
* ExpressionInCpsMethodNotSerializable
* ForbiddenCallInCpsMethod
* OverridesNotNonCps
* ParameterOrReturnTypeNotSerializable

* NotAllFieldsOfSerializableClassSerializable (serialization ruleset)


chrismair:
Wow! Thanks for this contribution! This is a cool new direction/capability for CodeNarc.

This is a HUGE PR -- much, much bigger than I am comfortable with. Can you submit this as one PR per rule -- one at a time? For instance, just that one `CpsCallFromNonCpsMethodRule` is a very large PR just by itself.

I would suggest making/converting this PR to an issue, and then link the multiple PRs to it.

Some quick initial comments:
 - Adding that new `groovy-cps` dependency is concerning and `https://repo.jenkins-ci.org/public/` repo too. Every new dependency is a headache down the road when it needs to be updated or moved or is no longer supported. Can it be avoided or simulated? I know using the SEMANTIC_ANALYSIS phase may make that difficult.
 - Update rule docs to say 'Since CodeNarc 3.5.0'
 - Probably worth including a link to the Jenkins CPS doc(s) in the ruleset doc, e.g. something like https://www.jenkins.io/doc/book/pipeline/cps-method-mismatches/

You have obviously put a lot of effort into the code quality and test coverage, and I do appreciate that a lot.

danielzaenker:
Thanks for the quick reply!

Sure, I can break down the PR into multiple PRs! I'm not sure if I can convert a PR into an issue but I can create a new one if that's not possible. 

The `groovy-cps` dependency is used for the `NonCPS` annotation only and made testing a little bit easier. I understand that an additional repo and dependency just for the tests is not ideal. I'll try to remove it. It should be possible to use a dummy annotation or mock some method calls instead.

By "ruleset doc" do you mean `docs/codenarc-rules-jenkins.md`? I already linked the Jenkins CPS docs if it made sense for a rule in there. Or do you mean just include an additional link at the top of the document?

danielzaenker:
I created a new issue (#764) and opened separate PRs for all Rules. Common changes are in seperate commits and have to be merged only once and all other PRs rebased afterwards.

chrismair:
Thanks. I'll take a look this weekend.

Issue #764, opened 2024-02-21
add new jenkins ruleset


danielzaenker:
Issue to track PRs for new jenkins ruleset.
See also #763

chrismair:
Please add some tests for `JenkinsUtil` and `SerializationUtil`.

Can we default all these rules to only apply to files named "Jenkinsfile", by setting the rule _applyToFileNames_ property? 
[CodeNarc - Configuring Rules](https://codenarc.org/codenarc-configuring-rules.html#standard-properties-for-configuring-rules)
Users could always override that if they used different naming conventions. But that would reduce the chances of applying the rules to other Groovy files.

chrismair:
I think another option for allowing testing with the @NonCPS annotations is to make the `JenkinsUtil` **NON_CPS_ANNOTATION** field non-final (and presumably then called _nonCpsAnnotation_). Then the tests could define a local placeholder version of the annotation and use that instead.

Or another option would be to define a do-nothing `com.cloudbees.groovy.cps.NonCPS` annotation class within the _test_ source folder.

danielzaenker:
Good idea, thanks. This seems a little bit less hacky than overriding the method. I'll have a look at it but it will take some time. I won't be able to work on it again for a few weeks.

chrismair:
No hurry. We are still months away from the next release.

In addition to the feedback I gave above and on the individual PRs, please also review the "codenarc-rules-jenkins.md" .. that was the only real problematic part of merging in these multiple PRs, and it is very possible the text got messed up. Or just replace it with your latest copy.

danielzaenker:
Thanks for the feedback! It should all be implemented in the PRs so I'm closing the issue. Otherwise just let me know

PR #772, opened 2024-02-21
add NotAllFieldsOfSerializableClassSerializableRule

danielzaenker:
See #764


src/main/groovy/org/codenarc/rule/serialization/NotAllFieldsOfSerializableClassSerializableRule.groovy
```diff
@@ -0,0 +1,47 @@
...
+import org.codenarc.rule.AbstractAstVisitor
+import org.codenarc.rule.AbstractAstVisitorRule
+import org.codenarc.util.AstUtil
+
+/**
+ * All fields of a class that implements Serializable should also implement it
+ *
+ * @author Daniel Zänker
+ */
+class NotAllFieldsOfSerializableClassSerializableRule extends AbstractAstVisitorRule {
```
chrismair, line 29: We could consider shortening this rule name. Perhaps `NonSerializableFieldInSerializableClass`.

chrismair (approved):

src/main/groovy/org/codenarc/rule/serialization/NotAllFieldsOfSerializableClassSerializableRule.groovy
danielzaenker: Renamed it in #779

PR #777, opened 2024-03-26
Speed up UnusedImportRule

dragos:
I noticed in our project (90KLoc of Groovy) that this rule was taking about 35% of all the running time. Using a profiler, I noticed that almost all the time was spent in regex matching.

The better long-term fix would be to rewrite this rule to use the AST, which already breaks down the source code into a meaningful structure, but that would require more work. What I did instead:

- pull regular expressions out of the hot loop and make sure they are compiled only once
- add a fast-path check where, if a line does not contain the import alias, consider there no matches. This takes into account that the regex has several alternatives, and it is the slowest when there are no matches (since it needs to try all alternatives)

This brings down the execution time on our project by about 20%, and the UnusedRule takes only 10% of the total time. This is still significantly more than other rules, but much better than before.

Before
Screenshot 2024-03-26 at 15 00 59

After:
Screenshot 2024-03-26 at 12 49 28


chrismair:
This is great. Thanks very much for that contribution.

Yeah, it is probably worth revisiting that (old) rule and see if we can use the AST for the import references as well.

Issue #785, opened 2024-09-22
Add support for GitLab's Code Quality service


UlrichEckhardt:
Hi!

GitLab's [Code Quality](https://docs.gitlab.com/ee/ci/testing/code_quality.html) is a part of its merge request / code review facilities. In order to use it, you upload reports (e.g. like the one generated by CodeNarc) to it and those reports are then presented to the reviewers in a merge request. This takes a very pragmatic approach to this topic: Instead of enforcing strict rules on the acceptable or inacceptable flaws in the code, it gives an overview of new (but also fixed) flaws to help the reviewer. I call this pragmatic, because setting up a ruleset in and of itself is an enormous task, and can easily stall or cripple adoption of a linter.

I'm currently working on a report generator for this. Actually, it's pretty simple and I'm almost finished, just need to clean up and test a bit. Just wanted to present this here up front, perhaps to answer questions and receive suggestions on the matter.

Greetings!

Uli

PR #786, opened 2024-09-24
GitLab Code Quality output

UlrichEckhardt:
This implements the feature I already announced in #785.

Demonstrating that it works is a bit difficult, because it requires a complete context switch from GitHub here to GitLab. I've set up an [example project on GitLab](https://gitlab.com/UlrichEckhardt/CodeNarc/-/merge_requests/1), actually a CodeNarc fork, in order to test this implementation, but I'm not sure whether this is really good to demo it to others. Please ask if there are things that are unclear.

The UI there displays this currently:
image. The comparison ("6 new findings") is with a dummy report. Normally, that is compared with the report from the branch which you're trying to merge into, but that needs this report writer. To work around this chicken-and-egg problem, I just used a dummy report.

BTW: The first commit is just refactoring, adding a few type annotations, and technically independent of this PR, but it made the new feature easier to develop for me.

Greetings from Hamburg!

Uli


chrismair:
Thanks. I'll take a look.

chrismair:
Looks good so far.

Issue #790, opened 2025-01-21
Add option to exclude getters/setters from PublicMethodsBeforeNonPublicMethods and StaticMethodsBeforeInstanceMethods rule


peter-thomas-mgd:
Our coding conventions group getters and setters before all other public or private methods, including constructors. It would be nice to have an option in the PublicMethodsBeforeNonPublicMethods and StaticMethodsBeforeInstanceMethods rules to exclude getters and setters. Without this option we need to disable both of these rules, which misses other public, private and static method ordering issues.

chrismair:
If the getters and setters are public, what is the problem? Aren't they already expected first? Do you have some examples of method ordering that is problematic?

And I am also curious about the motivation/rationale for that convention. What makes getters/setters "more" important than other types of methods?

peter-thomas-mgd:
Additional context: we often have private getters/setters that we want to appear at the top of the class (similar to the rationale given in the code snippet). I should have shown that in my example. We are not able to do this with the current `PublicMethodsBeforeNonPublicMethods` rule implementation. We follow this convention in our front-end TypeScript code base and want to keep a consistent set of rules/conventions between the front-end and back-end linting systems.

chrismair:
I don't think I have encountered that convention before. Rather than extending those rules with something so specific, I'd be more inclined to support a more general exclusion mechanism. 

That would probably be:
`ignoreMethodNames` - Specifies one or more (comma-separated) method names that should be ignored (i.e., that should not cause a rule violation). The names may optionally contain wildcards (*,?).

So you could ignore get* and set* methods. PRs are welcome.

peter-thomas-mgd:
That would work nicely. It follows the same pattern as some of the other exclusion rules. If I have some spare cycles over the next few days I'll try to take this library for a spin and see what I can work up.

chrismair:
Sounds good. I'm planning for a release in a few weeks.

peter-thomas-mgd:
I've created a pull requests with the changes (https://github.com/CodeNarc/CodeNarc/pull/792).

chrismair:
PR #792 was merged. That added _ignoreMethodNames_ to **PublicMethodsBeforeNonPublicMethods** and **StaticMethodsBeforeInstanceMethods** rules. 

That will be available in CodeNarc 3.6.0

PR #792, opened 2025-01-26
Add ignore options to method order rules

peter-thomas-mgd:
Added `ignoreMethodNames` option to `PublicMethodsBeforeNonPublicMethodsRule` and `StaticMethodsBeforeInstanceMethodsRule`. Added new tests and updated `codenarc-rules-convention.md` documentation. See [issue #790](https://github.com/CodeNarc/CodeNarc/issues/790) for context and rationale.

### DIFF
--- a/docs/codenarc-rules-convention.md
+++ b/docs/codenarc-rules-convention.md
@@ -474,6 +474,10 @@ Example of violations:
 
 Enforce that all public methods are above protected and private methods.
 
+| Property                    | Description            | Default Value    |
+|-----------------------------|------------------------|------------------|
+| ignoreMethodNames           | Specifies one or more (comma-separated) non-public method names that should be ignored (i.e., that should not cause a rule violation). The names may optionally contain wildcards (*,?).  | `null`  |
+
 Example of violations:
 
 ```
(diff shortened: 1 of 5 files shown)


chrismair:
The _ignoreMethodNames_ flags should completely ignore those methods. I would not expect those methods to cause violations anywhere. I can envision other scenarios than yours where a user would want to exclude certain methods from consideration.

peter-thomas-mgd:
The main goal of this work is to allow some limited exceptions but still enforce consistent ordering semantics. Having methods from the _ignoreMethodNames_ list appear anywhere in the file, with the potential to break the other ordered groupings was not the goal. For example, if the code is refactored, we still want to catch an instance of the ignored method unintentionally appearing in a random location elsewhere in the code.  
  
I wrote the original logic to ignore the list items completely, but then I realized there were edge cases that we should probably handle. The idea being that if someone is going to take the time to define entries in the ignore list, they are likely trying to implement a consistent ordering standard of some kind.
  
To turn off **all** checks for the ignored methods, what are your thoughts on introducing an option which allows for this behavior (e.g. `disableAllChecksOnIgnored`).

peter-thomas-mgd:
Not a hill worth dying on. But in case you wanted to understand how much more pervasive a "thing" this is, have a look at this:
https://typescript-eslint.io/rules/member-ordering/. There's a fairly large community of practice coding around these principles...

peter-thomas-mgd:
Pull request updated to fully ignore the entries in the ignore list.

chrismair:
Thanks for making those changes.

Interesting .. I was expecting a more narrow interpretation for _ignoreMethodNames_, where it would specify the names of the public methods (for **PublicMethodsBeforeNonPublicMethodsRule**) that would otherwise cause a violation. Your implementation includes that, but also ignores the  methods totally, and those named methods do not contribute to another (unnamed) public method causing a violation. That is just as reasonable an interpretation/expectation. And maybe that is even better. I need to think on that a bit.

In any case, I'll want to add some tests for naming the public methods, too, and having those violations being skipped, and also adjust the rule documentation to make clear that the _ignoreMethodNames_ also applies to the named public method. Thanks.

PR #797, opened 2025-07-14
Additional `PropertiesFileRuleSetConfigurer.configure(..)` log aiding debugging.

dlehammer:
This (or similar) debug logging would've been very helpful in teasing out the root cause of https://github.com/gradle/gradle/issues/34239 , when attempting to reason about state versus output for the `getActualPropertiesFile(..)` branches 🤓

### DIFF
--- a/src/main/groovy/org/codenarc/ruleset/PropertiesFileRuleSetConfigurer.groovy
+++ b/src/main/groovy/org/codenarc/ruleset/PropertiesFileRuleSetConfigurer.groovy
@@ -76,7 +76,11 @@ class PropertiesFileRuleSetConfigurer implements RuleSetConfigurer {
             }
         }
         catch(IOException e) {
-            LOG.info("RuleSet configuration properties file [$actualPropertiesFilename] not found.")
+            if (LOG.debugEnabled) {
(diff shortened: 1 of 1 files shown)


src/main/groovy/org/codenarc/ruleset/PropertiesFileRuleSetConfigurer.groovy
```diff
@@ -76,7 +76,11 @@ class PropertiesFileRuleSetConfigurer implements RuleSetConfigurer {
             }
         }
         catch(IOException e) {
-            LOG.info("RuleSet configuration properties file [$actualPropertiesFilename] not found.")
```
chrismair, line 79: This code only gets executed once and so I'm not too concerned about performance.

What do you think about something like this instead:
```
String originalConfig = propertiesFilename ? "; propertiesFilename=$propertiesFilename" : ""
LOG.info("RuleSet configuration properties file [$actualPropertiesFilename] not found" + originalConfig + ".")
```

dlehammer: Hi @chrismair ,

Thank you for your input 🙏 

I've taken a stab at adressing your feedback in d2ed225e8a3a33d79622c982625b21961478ec86, while preserving the `GString` instance, allowing for lazy evaluation runtime 🤓

chrismair:
Thank you for that contribution!

PR #798, opened 2025-07-21
Docs indent under label and adjustments

dauer:


chrismair:
Thanks very much for that contribution. All those incorrect escaping you fixed are leftovers from switching the docs to markdown several years ago -- thanks for that.

Issue #801, opened 2025-10-13
Docs: some broken URLs for the "Basic" ruleset


immutability:
Hello - in the documentation file codenarc-rules-basic.md, under ClassForName rule, there are two links to related blog posts, which are both broken now. 

For the first one pointing to blog.bjhargrave.com, I only found a working copy in the web archive at https://web.archive.org/web/20111107184503/http://blog.bjhargrave.com/2007/09/classforname-caches-defined-class-in.html.

For the 2nd one, this seems to be the current URL
https://blog.osgi.org/2011/05/what-you-should-know-about-class.html

Could someone please fix these? Thank you! :)

PR #802, opened 2025-10-14
Issue 801: Fix broken URLs for the "Basic" ruleset

dauer:
Fixed broken links as reported in issue 801:
  https://github.com/CodeNarc/CodeNarc/issues/801

Both blog entries had been moved but they stil existed


chrismair (commented):
Thanks very much for that fix!

chrismair (approved):

PR #821, opened 2026-03-21
Add `SpockUseVerifyEach` rule

### LINKED ISSUE #820: New Rule: `SpockUseVerifyEach` — enforce `verifyEach` over iteration with assertions
## Description

I propose a new rule `SpockUseVerifyEach` that detects `.every`, `.each`, `.eachWithIndex`, and `.forEach` calls containing assertions in Spock specifications, and suggests replacing them with Spock's `verifyEach` for better per-item failure diagnostics.

## Motivation

When iterating a collection to assert on each item, common patterns like:

```groovy
then:
list.every { it.field == value }
// or
list.each { assert it.field == value }
```

...fail fast on the first failure and provide poor diagnostics. Spock 2.4 introduced `verifyEach(collection) { ... }` which collects failures for each item individually, giving much better test output. This rule nudges users toward that superior pattern.

## Detection targets

The rule would flag these method calls when they contain assertions (explicit or implicit):

| Method | In `then:`/`expect:` | In other blocks |
|---|---|---|
| `.every { booleanExpr }` | Flagged (implicit assertion) | Not flagged (just a boolean return) |
| `.each { assert ... }` | Flagged | Flagged (if `checkAllBlocks = true`) |
| `.eachWithIndex { assert ... }` | Flagged | Flagged (if `checkAllBlocks = true`) |
| `.forEach { assert ... }` | Flagged | Flagged (if `checkAllBlocks = true`) |

Helper methods within spec classes would be treated like "other blocks."

**Violation message:** `"Replace '<methodName>' with Spock's 'verifyEach' for better per-item failure diagnostics"`

## Configuration Properties

| Property | Type | Default | Description |
|---|---|---|---|
| `specificationSuperclassNames` | String | `'*Specification'` | Wildcard pattern for spec superclass |
| `specificationClassNames` | String | `null` | Wildcard pattern for spec class names |
| `checkAllBlocks` | boolean | `true` | When true, checks all blocks (not just `then`/`expect`) for explicit assertions in iteration |

## Refactoring note

Implementation would also extract a shared `SpockUtil` utility class from `SpockMissingAssertRule`, consolidating duplicated logic (spec class matching, boolean expression detection, Spock label constants) for reuse across Spock-related rules.

## Open question: enabled by default?

`verifyEach` was added in **Spock 2.4**. Since CodeNarc is a static analysis tool and doesn't have Spock on its classpath, it cannot detect the Spock version at runtime. Should this rule be **enabled by default**, requiring users on Spock < 2.4 to explicitly disable it? Or should it be **disabled by default**, requiring users to opt in?

Arguments for enabled by default: consistent with other CodeNarc rules; Spock 2.4+ is becoming the norm.
Arguments for disabled by default: avoids false positives for projects still on older Spock versions.

leonard84:
## Summary

Closes #820

Adds a new `SpockUseVerifyEach` rule that detects iteration patterns (`.every`, `.each`, `.eachWithIndex`, `.forEach`) containing assertions in Spock specifications and suggests replacing them with Spock 2.4's `verifyEach`. This improves test diagnostics by reporting all per-item failures instead of failing fast on the first one.

The rule handles two distinct contexts:
- **Implicit assertion blocks** (`then:`/`expect:`): detects `.every { booleanExpr }` (implicit boolean assertion) and `.each`/`.eachWithIndex`/`.forEach` with implicit or explicit `assert` statements
- **Other blocks** (`when:`/`setup:`/etc.): only flags iterations with explicit `assert` keywords (controlled by `checkAllBlocks`, enabled by default)

## Implementation notes

- Extracted shared Spock AST utilities (`SpockUtil`) from `SpockMissingAssertRule` to avoid duplication — both rules now share label detection, specification matching, and closure analysis logic
- `SpockUtil` also adds support for newer Spock block labels (`filter`, `combined`, `group`) and registers `verifyEach` as a method with implicit assertions
- 273-line test suite with parameterized tests covering positive cases, negative cases, `checkAllBlocks` toggle, and helper method scenarios

## AI Disclosure

This PR was created with AI assistance, but every line was reviewed by me, and the code was polished by hand.


leonard84:
The failure `CodeNarc/CodeNarc/src/main/java/org/codenarc/rule/AbstractAstVisitorRule.java uses or overrides a deprecated API.` seems unrelated to my changes.

chrismair:
Thanks for the contribution. I'll take a look at it.

chrismair:
I see a failure in `RunCodeNarcAgainstProjectSourceCodeTest`  on [Add `SpockUseVerifyEach` rule · CodeNarc/CodeNarc@f0c0bb0](https://github.com/CodeNarc/CodeNarc/actions/runs/23383388447/job/68026311534?pr=821)
.. did you run all tests against your branch?

src/main/groovy/org/codenarc/rule/junit/SpockMissingAssertRule.groovy
```diff
@@ -34,18 +30,15 @@ import org.codehaus.groovy.ast.stmt.WhileStatement
...
 /**
  * Spock treats all expressions on the first level of a then or expect block as an implicit assertion. However,
  * everything inside if/for/switch/... blocks is not an implicit assert, just a useless comparison (unless wrapped by a `with` or `verifyAll`).
- *Ø
+ *
  * This rule finds such expressions, where an explicit call to assert would be required. Please note that the rule might
  * produce false positives, as it relies on method names to determine whether an expression has a boolean type or not.
  *
- * @author Jean André Gauthier
+ * @author Jean Andre Gauthier
```
chrismair: Restore the special character in his name, unless there is a compelling reason to replace it.

src/main/groovy/org/codenarc/rule/junit/SpockUseVerifyEachRule.groovy
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Detects usages of .every, .each, .eachWithIndex, and .forEach with assertions
+ * in Spock specifications, suggesting Spock 2.4's verifyEach instead.
+ *
+ * @author Leonard Brünings
+ */
+class SpockUseVerifyEachRule extends AbstractAstVisitorRule {
+
+    String name = 'SpockUseVerifyEach'
+    int priority = 2
```
chrismair: Switch this to priority 3. We reserve 2 for actual problems/errors with the code; 3 is for optimizations/style issues like this one.

src/main/groovy/org/codenarc/rule/junit/SpockUseVerifyEachRule.groovy
```diff
@@ -0,0 +1,119 @@
...
+
+        boolean shouldReport = inImplicitAssertBlock
+            ? isImplicitAssertionOverIteration(methodName, closureArg)
+            : isExplicitAssertionOverIteration(methodName, closureArg)
+
+        if (shouldReport) {
+            addViolation(statement, "Replace '${methodName}' with Spock's 'verifyEach' for better per-item failure diagnostics")
+        }
+    }
+
```
chrismair, line 92: Remove extra blank line; 1 blank line is sufficient

src/main/groovy/org/codenarc/rule/junit/SpockUtil.groovy
```diff
@@ -0,0 +1,150 @@
...
+        return method != null && BOOLEAN_METHOD_PATTERNS.any { it -> method.value.toString().matches(it) }
+    }
+
+    static boolean isImplicitAssertBlock(String label) {
+        return label in LABELS_WITH_IMPLICIT_ASSERTIONS
+    }
+
+    static boolean isSpockFeatureMethod(MethodNode node) {
+        if (node.code instanceof BlockStatement) {
+            BlockStatement block = (BlockStatement) node.code
```
chrismair, line 97: Restore comments that were here and elsewhere in the original code

src/test/groovy/org/codenarc/rule/junit/SpockUseVerifyEachRuleTest.groovy
```diff
@@ -0,0 +1,334 @@
...
+/**
+ * Tests for SpockUseVerifyEachRule
+ *
+ * @author Leonard Brünings
+ */
+class SpockUseVerifyEachRuleTest extends AbstractRuleTestCase<SpockUseVerifyEachRule> {
+
+    private static String violationMessage(String methodName) {
+        "Replace '${methodName}' with Spock's 'verifyEach' for better per-item failure diagnostics"
+    }
```
chrismair: Move this private member below the public methods

leonard84: fixed a7cb53b

src/main/groovy/org/codenarc/rule/junit/SpockUtil.groovy
leonard84: fixed a7cb53b

src/main/groovy/org/codenarc/rule/junit/SpockUseVerifyEachRule.groovy
leonard84: fixed a7cb53b

src/main/groovy/org/codenarc/rule/junit/SpockUseVerifyEachRule.groovy
leonard84: fixed a7cb53b

src/main/groovy/org/codenarc/rule/junit/SpockMissingAssertRule.groovy
leonard84: fixed a7cb53b

leonard84:
Thanks for the review @chrismair, I've addressed all the comments.

> I see a failure in `RunCodeNarcAgainstProjectSourceCodeTest` on [Add `SpockUseVerifyEach` rule · CodeNarc/CodeNarc@f0c0bb0](https://github.com/CodeNarc/CodeNarc/actions/runs/23383388447/job/68026311534?pr=821) .. did you run all tests against your branch?

I ran `./gradlew test` and `check` and all tests passed for me for 
a7cb53b.

chrismair (approved):
Looks good. Thanks for the contribution!

PR #831, opened 2026-07-25
New rule OneTopLevelClass

rinsley:
This adds a new rule OneTopLevelClass to require exactly one top-level class/enum/interface per file. The rule is patterned off of and similar to ClassNameSameAsFilename but added to the convention ruleset instead, as this is more of a Java/Groovy convention than a naming issue. Classes inside scripts or other classes are not considered violations. Similar to some other convention rules, it has to be excluded from the RunCodeNarcAgainstProjectSourceCode.ruleset.


chrismair (approved):
Nice idea. Thanks for the contribution.

